PR #1, opened 2024-11-20
Cython Implementation and improved Python fib impl.

JHowell45:
You'll need to make sure you have python installed as in order for Cython to work you'll need to install Cython and Setuptools. This will be used for building the cython files for the tests. Ideally you'll want the latest, Python 3.13.x, but any python should work I think.


bddicken:
Thanks for the PR! I'll Make sure and go through it either this evening or tomorrow.

bddicken:
Looks like a bunch of extra files got added in to the PR? Would be great to add cython, but can you clean it up a bit first?

JHowell45:
Apologies I was messing about with pyx files. I'm not massively familiar with Cython, so for anyone looking at this I know we can improve the fibonacci method using .pyx and defining the fibonacci method as a cdef one. I'm just not able to 😂 

All the extra files should be cleaned up now 👍

bddicken:
I poked around, and not really sure if this is working correctly? It doesn't seem like `python3 ./pyx/code.py 40` is the right way to run this? For me, it's not producing the output and does not seems like its actually executing. I'm also not familiar with cython, unfortunately.

bddicken:
I really appreciate the effort you've put in here, but I am not going to include Cython for this next round. It's basically just taking python and compiling it to optimized C code, but we already have C in the benchmark. For python, I'm just going to include standard interpreter / jits such as Cpython and pypy.

FI-Mihej:
> I really appreciate the effort you've put in here, but I am not going to include Cython for this next round. It's basically just taking python and compiling it to optimized C code, but we already have C in the benchmark. For python, I'm just going to include standard interpreter / jits such as Cpython and pypy.

Allow me to provide some necessary clarification. Cython is a specialized Python-like language (similar to Nim or Julia) that doesn’t just compile into highly efficient, but non-Python-related C code. Instead, it creates and compiles a Python-compatible C module that operates on the same C object structures used internally by the Python interpreter.

In other words, it doesn’t compile the given benchmark into C code that works with a plain C array of integers. No, it compiles it into C code that creates dynamic array structures consisting of pointers to objects. These objects not only store the integers themselves but also include all the additional metadata required to describe a Python object (e.g., size, reference to the type structure, etc.). Essentially, it works with exactly the same structures that the CPython interpreter, written in C, operates on.

The only difference is that Cython eliminates certain parts of the infrastructure responsible for decoding bytecode and performing the repetitive operations required to ensure data integrity (as needed in a full-fledged interpreter). It retains only the minimum necessary infrastructure to safely execute the code.

In addition to this, due to its compiled nature, Cython provides slightly better cache locality when executing the resulting binary. This means that the processor experiences fewer cache misses while running such code, which, in turn, slightly speeds up its execution.

The performance improvement when writing a module in Cython ranges from 20% to several times faster compared to CPython. However, it naturally doesn’t reach the level of a program written in C or C++. Unless, of course, the C/C++ program also wraps every integer in its own data structure, guarantees array bounds checking, and works with all the typically optional features of the target operating systems (e.g., mandatorily storing incoming signals in a processing pool and periodically checking the state of that pool to ensure the ability for proper delayed signal handling), and so on. In such a case, of course, you could fairly compare the performance of code written in C/C++ with that of CPython, Cython, or PyPy. :)

PR #2, opened 2024-11-20
feat: add java

brunopacheco1:
Package name as `java` is not allow, hence the folder is called `jvm`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+*.class
+code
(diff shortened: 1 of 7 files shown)


bddicken:
Thanks for the PR! I'll Make sure and go through it either this evening or tomorrow.

bddicken (approved):
This is excellent, thanks for contributing.

PR #3, opened 2024-11-20
adding rust

arpadav:
added rust

requirements: [`cargo / rustup`](https://www.rust-lang.org/tools/install) 

a couple other changes, which i am happy to alter:

1. `gcc` instead of `clang`. there are benefits to both, but if anything it would be nice to test both maybe? happy to remove this, but `gcc` gave faster (barely) results compared to `clang` on my machine: `x86_64-linux-gnu`
2. added .gitignore files for the C and Go binaries
3. thoughts on `hyperfine` instead of the custom run function? i modified the `run.sh` to include, but it is commented out


Hoimar:
Can we get a sneak peek of what your results are looking like?

kfkonrad:
If you're suggesting hyperfine anyway, why not also use it to rank runtime between the benchmarks with `hyperfine "foo $NUM" "bar $NUM" ...`?

arpadav:
> Can we get a sneak peek of what your results are looking like?

I assume you mean for Rust, not the `gcc` -> `clang`:

For `loops` it was uneventful, around the same as the fastest ones. `fibonacci`, however, was about 20% faster than C, which was surprising.

I can run it on my personal machine and include specs when I get back home.

arpadav:
> If you're suggesting hyperfine anyway, why not also use it to rank runtime between the benchmarks with `hyperfine "foo $NUM" "bar $NUM" ...`?

Very true, I'll update this

bddicken:
Nice! Definitely need rust in the mix. I'll Make sure and go through it either this evening or tomorrow.

bddicken:
Awesome! Thanks for doing this. Could you please:

- Remove the changes you made for gcc / clang. We could add in gcc later, but I'd like to keep this PR just to rust.
- Revert the changes to `run.sh` and just add `run "Rust" "./rust/target/release/code 40"` for now?

Then also, rebase onto main so we can get it merged.

arpadav:
Some benchmarks @Hoimar 

Run using:

```bash
NUM=40

hyperfine \
  "./c/code $NUM" \
  "./go/code $NUM" \
  "node ./js/code.js $NUM" \
  "bun ./js/code.js $NUM" \
  "deno ./js/code.js $NUM" \
  "pypy ./py/code.py $NUM" \
  "./rust/target/release/code $NUM" \
  ```

I was very surprised when I saw Rust perform better than C in the Fibonacci test on my Linux server (benchmarked around 17% better), so I had to test it on a couple other machines. I find it very strange how different these results are across machines and architectures, and I think it might be a combination of `gcc` version being out-of-date as well as the Linux server being beefier. But the benchmarks are attached none-the-less
  
[fibonacci.md](https://github.com/user-attachments/files/17847598/fibonacci.md)
[loops.md](https://github.com/user-attachments/files/17847599/loops.md)
[machines.md](https://github.com/user-attachments/files/17847600/machines.md)

**Comment by arpadav** [comment]
@bddicken rebased, should be good to go

and FYI, the reason I used `hyperfine` is because this is the output I am getting with your function, which is not what I think you intended:

```bash
C = 1.54user
0inputs+0outputs
Go = 1.80user
0inputs+0outputs
Node = 1.59user
0inputs+0outputs
Bun = 1.56user
0inputs+0outputs
Deno = 1.61user
0inputs+216outputs
PyPy = 1.61user
0inputs+0outputs
Rust = 1.55user
0inputs+0outputs
```

I am on Ubuntu 20.04.6 LTS, and running `time` and piping it to `/tmp/o` results in the following text that the `run` fn is trying to parse:

```
1.55user 0.00system 0:01.55elapsed 99%CPU (0avgtext+0avgdata 2100maxresident)k
0inputs+0outputs (0major+106minor)pagefaults 0swaps
```

Just wanted to let you know, in case anyone else had this issue

bddicken:
@arpadav Thanks for the detailed explainer. I was running this on macOS Sequoia, which probably accounts for the difference. Maybe we can look into hyperfine as a separate addition / pr? I just like to keep them isolated.

PR #4, opened 2024-11-21
add rust implementation for loop benchmark

saleem-mirza:
The code in rust is optimized using rayon https://docs.rs/rayon/latest/rayon/

Code implementation will take full advantage of modern multicore cpus. On my test machine, execution output is as below but your milage may vary.

```
time ./target/release/rust 100
4952670

real	0m0.076s
user	0m0.994s
sys	0m0.018s
```

Just for reference, C implementation's output

```
time ./cmain 100
4956807

real	0m0.741s
user	0m0.729s
sys	0m0.011s
```

### DIFF
--- a/clean.sh
+++ b/clean.sh
@@ -1,2 +1,4 @@
 rm c/code
 rm go/code
+cargo clean --manifest-path rust/Cargo.toml 
+
--- a/compile.sh
+++ b/compile.sh
@@ -1,2 +1,4 @@
 clang -O3 c/code.c -o c/code
 go build -o go/code go/code.go
+cargo build --release --manifest-path rust/Cargo.toml
+
--- a/loops/rust/.gitignore
+++ b/loops/rust/.gitignore
@@ -1,0 +1,3 @@
+.vscode
+Cargo.lock
+target/**
--- a/loops/rust/Cargo.toml
+++ b/loops/rust/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "rust"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+rand = "0.8.5"
+rayon = "1.10.0"
+
+[profile.release]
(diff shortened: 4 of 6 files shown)


bddicken:
Interesting! I'm not a rust-ite, but it does look like you're introducing "extra" logic here to handle parallelism. Technically, we could do that in all of the other languages too and see a big performance boost, but have not (at least as of now). I think it would be better to either all be single-thread or all be parallel.

arpadav:
spot on @bddicken 

the `rayon` crate performs a divide-and-conquer approach to parallelism. it would be nice if there is a different benchmark(s) for parallelism, since I agree that the `loops` benchmark was intentionally meant to be naively single-threaded

bddicken:
Ok cool. Well, I think we should keep the `loops` directory non-parallel. If you'd like to add a `parallel` directory as a new benchmark you're welcome to, but I'm not sure who else is interested in getting all the other languages up to speed.

saleem-mirza:
agreed, I didn't paid much attention. However, here is parallelized version of C code using openmp :)  enjoy

```
#include "stdio.h"
#include "stdlib.h"
#include "stdint.h"
#include "omp.h"

int main (int argc, char** argv) {
  int u = atoi(argv[1]);               // Get an input number from the command line
  int r = rand() % 10000;              // Get a random integer 0 <= r < 10k
  int32_t a[10000] = {0};              // Array of 10k elements initialized to 0
  
  #pragma omp parallel for
  {
    for (int i = 0; i < 10000; i++) {    // 10k outer loop iterations
        a[i] = a[i] + j%u;               // Simple sum
      }
      a[i] += r;                         // Add a random value to each element in array
    }
  }
  printf("%d\n", a[r]);                  // Print out a single element from the array
}

```

saleem-mirza:
closing PR, since it is no more relevant

PR #7, opened 2024-11-21
add-r

paladinic:
added R an implementation for loops and fibonacci


bddicken:
Nice, thanks for the PR! The implementation of the loops code does not match what the other programs do. Can you get it to match please?

bddicken:
NICE

PR #8, opened 2024-11-21
Small PHP optimisation for the loop code

JHowell45:
It's nearly always faster to use += over indexing the array again. Also took a look at SplFixedArray, even though it uses less memory and seems to be faster in certain circumstances, the overhead of creating the array not only undoes any performance gain, but actually increases the overall runtime.


bddicken:
Nice!

PR #9, opened 2024-11-21
Adding Dart

alvynfash:


bddicken:
Thank you for the contribution, it would be cool to add dart.

However, the code implementations do not match the other programs in the benchmark. For example, your fibonaccii function uses an iterative solution rather than recursive. Also, the loops test does not use a nested loop. If you get the programs to match up more precisely, we can get this added in.

bddicken:
Again thanks for the contribution, but Im going to go with https://github.com/bddicken/languages/pull/13 over this one as the code is closer to being correct and you have not responded.

PR #11, opened 2024-11-21
feat: add v fibonacci

1ethanhansen:


compile.sh
```diff
@@ -1,4 +1,5 @@
 clang -O3 c/code.c -o c/code
 go build -o go/code go/code.go
 javac jvm/code.java
-RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Cargo.toml --release 
\ No newline at end of file
+RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Cargo.toml --release 
+v v/code.v -o v/code
```
enghitalo, line 5: v -prod v/code.v -o v/code

to run with GCC instead of TCC

bddicken:
Thanks for the contribution! Not going to merge this as of now. It only has one of the two benchmark programs.

bddicken:
Have not heard back, closing out.

PR #15, opened 2024-11-25
kotlin jvm, js and native

programadorthi:
Bring kotlin JVM, JS and Native.

JVM is working as previosly.
Native needs [download and install the Kotlin Native Compiler.](https://kotlinlang.org/docs/native-command-line-compiler.html#download-and-install-the-compiler)
JS almost of the time come with JVM compiler but missing the lib folder that contains que `kotlin-stdlib-js.klib`. So a [manual install](https://kotlinlang.org/docs/command-line.html#manual-install) would be better.
After that open `compile.sh` and set `KOTLIN_HOME` to Kotlin directory.


compile.sh
```diff
@@ -2,6 +2,31 @@ clang -O3 c/code.c -o c/code
 go build -o go/code go/code.go
 javac jvm/code.java
 RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Cargo.toml --release
-kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
-#kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+kotlinc-jvm -include-runtime kotlin/jvm/code.kt -d kotlin/jvm/code.jar
+kotlinc-native kotlin/native/code.kt -o kotlin/native/code
```
dayanruben, line 7: For better runtime performance you can add `-opt` argument. For reference: [docs](https://kotlinlang.org/docs/compiler-reference.html#opt).

PR #16, opened 2024-11-25
Add Haskell benchmarking set

ChAoSUnItY:
As titled, currently I'm using list comprehension syntax to build up a loop, might not be the best solution, but Glasglow Haskell Compiler (a.k.a. GHC) should be powerful enough to optimize into loop.

## Building Notes

Considering using GHCup to install GHC, following by installing `System.Random` package with command:
```bash
$ cabal install --lib random
```
and you should be able to run it with script without further issue. The reason why not using cabal or stack management is that it generates executable inside output folder with system dependent, it's kind of tricky to retreive the exact location of executable without lots of machine information retrievement.


haskell-monad:
```haskell
import System.Environment (getArgs)
import System.Random (randomRIO)
import Data.Vector.Unboxed (Vector, generate, (!))
import qualified Data.Vector.Unboxed as V

-- Compute a single element in the array
computeElement :: Int -> Int -> Int -> Int
computeElement u r i = foldl' (\acc j -> acc + (j `mod` u)) 0 [0..99999] + r

main :: IO ()
main = do
    args <- getArgs
    let u = read (head args) :: Int           -- Get an input number from the command line
    r <- randomRIO (0, 9999)                  -- Get a random number 0 <= r < 10k
    let a = V.generate 10000 (computeElement u r) -- Generate the array using Vector
    print (a ! r)                             -- Print out a single element from the array

--  ./ghc -O2 -threaded -rtsopts main.hs
```


```haskell
import Control.Parallel.Strategies (parList, rdeepseq, using)

main :: IO ()
main = do
    args <- getArgs
    let u = read (head args) :: Int
    r <- randomRIO (0, 9999)
    let a = V.fromList $ computeElement u r <$> [0..9999]
                 `using` parList rdeepseq
    print (a ! r)
```

ChAoSUnItY:
I've tested and timed your code under same compilation flags, but I don't think the vector approach has significant performance benefit than the list comprehension approach:

`list comprehension approach`:
```bash
$ time ./code 1000
49856988

________________________________________________________
Executed in   12.00 millis    fish           external
   usr time    2.13 millis  100.00 micros    2.03 millis
   sys time    0.16 millis  162.00 micros    0.00 millis
```

`vector approach`:
```bash
$ time ./code 1000
49955954

________________________________________________________
Executed in   12.08 millis    fish           external
   usr time    2.67 millis  124.00 micros    2.55 millis
   sys time    0.20 millis  197.00 micros    0.00 millis
```

Additionally, while other languages didn't use parallelism to further optimize their program, I think Haskell shouldn't optimize program with parallelism in this case.

compile.sh
```diff
@@ -5,3 +5,4 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+ghc -O2 haskell/code.hs -o haskell/code
```
kimminss0, line 8: Performance might be improved with additional `-fllvm` flag, which uses LLVM backend for compilation. It really matters in some cases according to the [documentation]. Here is my personal experiment: [link][benchmark]

Note: currently supported LLVM version is 13, 14, 15; compilation would fail otherwise.

```sh
if ghc -O2 -fllvm haskell/code.hs -o haskell/code; then
    true
else
    echo "ghc: cannot compile with llvm backend; fallback to use default backend"
    ghc -O2 haskell/code.hs -o haskell/code
fi
```

Or one liner:

```sh
ghc -O2 -fllvm haskell/code.hs -o haskell/code || { echo "ghc: cannot compile with llvm backend; fallback to use default backend"; ghc -O2 haskell/code.hs -o haskell/code; }
```

[documentation]: https://downloads.haskell.org/ghc/latest/docs/users_guide/codegens.html
[benchmark]: https://github.com/kimminss0/haskell-playground/tree/92fc63f15a41fcdd5c9260a9b4ed45554d2cc9a2/fibonacci/benchmark

kimminss0: Tested on fibonacci:

with llvm 14:
```
% \time  ./code 40
165580140
        0.72 real         0.70 user         0.00 sys
```

without llvm; using default backend:
```
% \time  ./code 40
165580140
        0.93 real         0.91 user         0.00 sys
```

ChAoSUnItY: Thanks for your suggestion! I have adopted it in latest commit.

bddicken:
Thanks!

PR #18, opened 2024-11-25
Swift benchmarks

vishrutkmr7:
## macOS Setup
Swift benchmarks run natively on macOS with Xcode
- Requirements:
  - Xcode installed
  - Terminal access

## Ubuntu Setup
**Installation:**
1. Download Swift toolchain from [swift.org](https://www.swift.org/install/linux/ubuntu/24_04/#latest)
2. Verify setup:
```bash
swift --version
```


nervenes:
What's with the prerequisites? All you need is a swift toolchain from [swift.org](https://swift.org).

Change the optimization flag to ``-Ounchecked`` to match with rustc and clang optimizations.

nervenes:
Also, remove the unnecessary ``import Foundation``

loops/swift/code.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+let u = Int(CommandLine.arguments[1])!  // Get input number from command line
+let r = Int.random(in: 0..<10000)  // Get random number 0 <= r < 10k
+var a = Array(repeating: 0, count: 10000)  // Array of 10k elements initialized to 0
+
+for i in 0..<10000 {  // 10k outer loop iterations
+    for j in 0..<100000 {  // 100k inner loop iterations
+        a[i] = a[i] + j % u  // Simple sum
```
nervenes, line 9: why are you not using ``+=`` instead, which you seem to do in the js code you modified.

fibonacci/py/code.py
```diff
@@ -3,12 +3,8 @@
...
-  if (n == 1):
-    return 1
-  return fibonacci(n-1) + fibonacci(n-2)
+  return 1 if (n == 1) else fibonacci(n-1) + fibonacci(n-2)
 
 u = int(sys.argv[1])
-r = 0
-for i in range(u):
-  r += fibonacci(i)
+r = sum(fibonacci(i) for i in range(u))
```
nervenes, line 9: why are you changing this? your change makes the python code semantically different to all samples in the other languages.

vishrutkmr7: Just ran a linter that made those changes. Will remove the changes made to py and js files. Sorry

loops/swift/code.swift
```diff
@@ -0,0 +1,12 @@
+let u = Int(CommandLine.arguments[1])!  // Get input number from command line
+let r = Int.random(in: 0..<10000)  // Get random number 0 <= r < 10k
+var a = Array(repeating: 0, count: 10000)  // Array of 10k elements initialized to 0
```
nervenes, line 3: seems like rust uses ``usize`` so might as well
```swift
let u = UInt(CommandLine.arguments[1])!
let r = UInt.random(in: 0..<10000)
var a: [UInt] = Array(repeating: 0, count: 10000)
```

fibonacci/swift/code.swift
```diff
@@ -0,0 +1,14 @@
+func fibonacci(_ n: Int) -> Int {
+    if n == 0 { return 0 }
+    if n == 1 { return 1 }
+    return fibonacci(n - 1) + fibonacci(n - 2)
+}
+
+let u = Int(CommandLine.arguments[1])!
+var r = 0
```
nervenes, line 8: same here
```swift
func fibonacci(_ n: UInt) -> UInt {
    if n == 0 { return 0 }
    if n == 1 { return 1 }
    return fibonacci(n - 1) + fibonacci(n - 2)
}

let u = UInt(CommandLine.arguments[1])!
var r = 0
```

compile.sh
```diff
@@ -5,3 +5,4 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+swiftc swift/code.swift -O -o swift/code
```
nervenes, line 8: ``swiftc swift/code.swift -Ounchecked -o swift/code``

nervenes (changes requested):

loops/swift/code.swift
```diff
@@ -0,0 +1,12 @@
...
+
+for i in 0..<10000 {  // 10k outer loop iterations
+    for j in 0..<100000 {  // 100k inner loop iterations
+        a[i] += j % Int(u)  // Simple sum
+    }
+    a[i] += Int(r)  // Add random value to each element
+}
+
+print(a[Int(r)])  // Print single element from array
```
nervenes, line 12: actually, probably better to do
```swift
let u = UInt(CommandLine.arguments[1])!
let r = UInt.random(in: 0..<10000)
var a: [UInt] = Array(repeating: 0, count: 10000)

for i in 0..<10000 {
    for j in 0..<10000 {
        a[i] += UInt(j) % u
    }
    a[i] += r
}

print(a[Int(r)])
```

vishrutkmr7: That's what it is, right now, right?

nervenes (approved):

loops/swift/code.swift
```diff
@@ -0,0 +1,12 @@
+let u = UInt(CommandLine.arguments[1])!  // Get input number from command line
+let r = UInt.random(in: 0..<10000)  // Get random number 0 <= r < 10k
+var a: [UInt] = Array(repeating: 0, count: 10000)  // Array of 10k elements initialized to 0
+
+for i in 0..<10000 {  // 10k outer loop iterations
+    for j in 0..<100000 {  // 100k inner loop iterations
+        a[i] += UInt(j) % u  // Simple sum
+    }
```
SeanRobinson159, line 8: Does type casting inside of the loop impact performance? Would it be better to change the for loop declaration to iterate over `UInt`s?

```suggestion
    for j in 0..<UInt(100000) {  // 100k inner loop iterations
        a[i] += j % u  // Simple sum
    }
```

vishrutkmr7:
@nervenes did some research. I think Int and UInt will have negligible differences in performance but the type casting is unnecessary and might affect performance. Should I revert back to Int?

loops/swift/code.swift
vishrutkmr7: exactly my point here https://github.com/bddicken/languages/pull/18#issuecomment-2499392211

SeanRobinson159: Oh I must have started my review before you posted that comment. Sounds like we are in agreement then!

SeanRobinson159 (approved):
The two JS files probably shouldn't change, since this is a Swift PR.

nervenes (approved):

MihaelIsaev:
Changes in the JS files seem to be related to the linter and should be reverted to the original

vishrutkmr7:
Will revert the changes made by the linter to JS files

nervenes:
@vishrutkmr7 so looking at the C and Rust code, both are using stack based arrays and both compilers omit safety checks by default in the generated asm for release, in Swift we dont have a stack based array yet (there is a proposal for it though) so we're using the dynamic and heap allocated ``Array`` type, but due to the nature of mutating it in the for loop the Swift generated asm with ``-Ounchecked`` still does a ``Swift._endMutation()`` check which is what made the swift code slower than the C and Rust code, both the Rust and C examples take around 0.50-0.53s to run on my end, and Swift were hovering over 0.70s. I think it would be fair to drop down to the unsafe api to allocate on the stack and bypass the safety check:
```swift
let u = UInt32(CommandLine.arguments[1])!
let r = UInt32.random(in: 0..<10000)

withUnsafeTemporaryAllocation(of: UInt32.self, capacity: 10000) { a in
    a.initialize(repeating: 0)
    
    for i in 0..<10000 {
        for j in 0..<100000 as Range<UInt32> {
            a[i] += j % u
        }
        
        a[i] += r
    }
    
    print(a[Int(r)])
}
```
this code now takes around 0.50-0.53s just like C and Rust.

loops/swift/code.swift
```diff
@@ -0,0 +1,12 @@
+let u = Int(CommandLine.arguments[1])!  // Get input number from command line
```
weissi, line 1: Please avoid top-level code in Swift, that doesn't benefit from all the optimisations. In this case, it'll be really noticeable. Ideally write

```swift
@main
struct Main {
    static func main() {
        let u = Int(CommandLine.arguments[1])!  // Get input number from command line
        let r = Int.random(in: 0..<10000)  // Get random number 0 <= r < 10k
        var a: [Int] = Array(repeating: 0, count: 10000)  // Array of 10k elements initialized to 0

        for i in 0..<10000 {  // 10k outer loop iterations
            for j in 0..<100000 {  // 100k inner loop iterations
                a[i] += j % u  // Simple sum
                // [if you wanted to disable overflow check here, you can optimise this with] a[i] = a[i] &+ (j % u)  // Simple sum
            }
            a[i] += r
            // [if you wanted to disable overflow check here, you can optimise this with] a[i] = a[i] &+ r  // Add random value to each element
        }

        print(a[r])  // Print single element from array
    }
}
```

Compilation: `swiftc -O -parse-as-library -o swift/executable swift/code.swift`

fibonacci/swift/code.swift
```diff
@@ -0,0 +1,14 @@
...
+
+for i in 1..<u {
```
weissi, line 10: here too, please use `@main` to avoid top-level code which doesn't benefit from (all) the optimisations.

loops/swift/code.swift
nervenes: TIL wasn't aware of this, this seems to yield a slightly faster result for me:
```swift
@main
struct Main {
    static func main() {
        let u = Int(CommandLine.arguments[1])!
        let r = Int.random(in: 0..<10000)

        withUnsafeTemporaryAllocation(of: Int.self, capacity: 10000) { a in
            a.initialize(repeating: 0)
            
            for i in 0..<10000 {
                for j in 0..<100000 {
                    a[i] += j % u
                }
                
                a[i] += r
            }
            
            print(a[r])
        }
    }
}
```

vishrutkmr7:
I never knew I'd learn this much from a simple looping/Fibonacci code. Thanks, everyone!

nervenes:
This is the fastest i could make it @vishrutkmr7 @weissi :
```swift
@main
struct Main {
    static func main() {
        let u = Int(CommandLine.arguments[1])!
        let r = Int.random(in: 0..<10000)

        withUnsafeTemporaryAllocation(of: Int.self, capacity: 10000) { a in
            a.initialize(repeating: 0)
            
            for i in 0..<10000 {
                for j in 0..<100000 {
                    a[i] += j % u
                }
                
                a[i] += r
            }
            
            print(a[r])
        }
    }
}
```
``swiftc -O -parse-as-library -Xcc -funroll-loops -Xcc -march=native -Xcc -ftree-vectorize -Xcc -ffast-math swift/code.swift -o swift/code``

Please apply these compiler flags as well @vishrutkmr7

vishrutkmr7:
@bddicken I guess we've optimized for Swift to the best we can!

nervenes (approved):

MihaelIsaev (approved):

nervenes:
actually, it does seem like these compiler flags ``-Xcc -funroll-loops -Xcc -march=native -Xcc -ftree-vectorize -Xcc -ffast-math`` arent really improving anything for these benchmarks, so i'd guess they're redundant and can be removed for simplicity instead, my bad!

fibonacci/swift/code.swift
```diff
@@ -0,0 +1,19 @@
+@main
+struct Main {
+    static func fibonacci(_ n: Int) -> Int {
+        if n == 0 { return 0 }
+        if n == 1 { return 1 }
```
nervenes, line 5: this seems to be faster for me:
```swift
if n <= 1 { return n }
```

nervenes (changes requested):

nervenes:
the fibonacci code is much faster with ``-Ounchecked`` but that slows down the other loop code, for fastest result for *both* with the same compiler flags seems to be this:
``swiftc -Ounchecked -parse-as-library``
```swift
@main
struct Main {
    static func main() {
        let u = Int32(CommandLine.arguments[1])!
        let r = Int32.random(in: 0..<10000)
        var a: [Int32] = Array(repeating: 0, count: 10000)

        for i in 0..<10000 {
            for j in 0..<100000 as Range<Int32> {
                a[i] += j % u
            }

            a[i] += r
        }

        print(a[Int(r)])
    }
}
```
```swift
@main
struct Main {
    static func fibonacci(_ n: Int) -> Int {
        if n <= 1 { return n }
        return fibonacci(n - 1) + fibonacci(n - 2)
    }

    static func main() {
        let u = Int(CommandLine.arguments[1])!
        var r = 0

        for i in 1..<u {
            r += fibonacci(i)
        }

        print(r)
    }
}
```

fibonacci/swift/code.swift
bddicken: But thats different than the other progs, so will leave as is.

bddicken:
This is very nice, thanks for the submission.

amomchilov:
>  but due to the nature of mutating it in the for loop the Swift generated asm with -Ounchecked still does a Swift._endMutation() check (to my understanding from observing the call stack with a profiler), which is what makes the swift code slower than the C and Rust code, both the Rust and C examples take around 0.50-0.53s to run on my end, and Swift were hovering over 0.70s.

This can also be avoided with `-enforce-exclusivity=unchecked`

loops/swift/code.swift
```diff
@@ -0,0 +1,21 @@
+@main
+struct Main {
+    static func main() {
+        let u = Int(CommandLine.arguments[1])!
+        let r = Int.random(in: 0..<10000)
+
+        withUnsafeTemporaryAllocation(of: Int.self, capacity: 10000) { a in
+            a.initialize(repeating: 0)
+
+            for i in 0..<10000 {
```
amomchilov, line 10: Since all the values are `0` from the `a.initialize(repeating: 0)` line, the first iteration can be skipped entirely

```suggestion
            for i in 1..<10000 {
```

Frizlab:
I have dumb question: will Swift appear [here](https://benjdd.com/languages/) at some point?

PR #19, opened 2024-11-25
Enable PHP's JIT

goetas:
As of PHP 8, JIT is available (many of the languages in the benchmark have JIT enabled by default)

With Jit enabled, on my Mac M3 php runs in 2.4s (previously was 9.6s).


goetas:
@lcobucci  I have tried it, and surprisingly, it performs worse. it takes ~11s. By reading the docs, it says that splfixedarray takes less memory but says nothing about performance .

https://www.php.net/manual/en/class.splfixedarray.php
> The advantage is that it uses less memory than a standard array.

lcobucci:
Interesting...

chuoke:
On my old laptop, using this command is slower than without it.

goetas:
@ilnytskyi my numbers were on php 8.4.1

goetas:
@chuoke  which php version do you have?

bddicken:
I have `PHP 8.3.13` on an M3 MBP, and for me, this does not seem to make a noticeable positive improvement for PHP. Am I doing something wrong?

bddicken:
I'll just include both a jit and non-jit option for PHP

goetas:
created https://github.com/bddicken/languages/pull/137

chuoke:
> @chuoke which php version do you have?

 v8.3.14.

BTW, v8.4.1 is the slowest of all versions currently on my laptop.

PR #20, opened 2024-11-25
add JuliaLang

marinlauber:
Julia examples for the `loop` and `fibonacci` benchmark in JuliaLang.

The simplest way to run these is to use `juliaup` to download the latest Julia release; see https://github.com/JuliaLang/juliaup.
ad then simply with
```bash
julia code.jl ARGS
```


loops/julia/code.jl
```diff
@@ -0,0 +1,12 @@
+let
+    a = zeros(Int,10^4)
+    u = parse(Int,ARGS[1])
+    r = only(rand(1:10^4,1))
+    @inbounds for i ∈ 1:10^4
+        @inbounds for j ∈ 1:10^5
+            a[i] = a[i] + u
```
silva96, line 7: You are not doing the module operation here, see the other languages.

silva96 (changes requested):

loops/julia/code.jl
marinlauber: You are right, I missed that!

loops/julia/code.jl
```diff
@@ -0,0 +1,12 @@
+let
+    a = zeros(Int,10^4)
+    u = parse(Int,ARGS[1])
+    r = rand(1:10^4)
+    @inbounds for i ∈ 1:10^4
+        @inbounds for j ∈ 1:10^5
```
serhii-havrylov, line 6: it should be `j ∈ 0:(10^5-1)` to correspond to other languages

fibonacci/julia/code.jl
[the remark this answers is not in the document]
bddicken: Yeah I'm pretty sure this is "cheating". We need to make it an actual function to be apples to apples. I'm gonna go ahead and merge and we can fix later.

PEZ:

* #371
* #410

PR #21, opened 2024-11-25
Add C# benchmarks

Odonno:


Agredo:
Does AOT make a difference? Mabe both would be great.

Odonno:
> Does AOT make a difference? Mabe both would be great.

Not the time to blink and code it that there's already a request for it. Damned.

So, here are some insights from I got, using the languages I can run ATM.

* Run on Debian (wsl2)
* Only run the "loops" test

```
C = 1.30
Node = 1.39
Bun = 1.32
C# = 0.15 (0.12 with native AOT)
```

I am not sure what to think of this result. Is it really that fast or did I made a mistake writing the program?
I can see the fastest in the list is Java but I can't run it unfortunately.

Odonno:
Ah yes. I missed a `0`...

amirvenus (approved):

loops/csharp/Program.cs
```diff
@@ -0,0 +1,14 @@
+﻿int u = int.Parse(args[0]);
+int r = Random.Shared.Next(0, 10000);
+int[] a = new int[10000];
+
```
RogerDerekWeiss, line 4: Assign `a` to a new variable of type `Span<int> span` and then mutate the span in the loop, its much faster.

tats-u: C and Rust use an array in the stack instead of the heap.
The size can be determined at the compilation phase so we can use Span in C#.

Odonno: Done.

However, it did not improve performance of the app. Memory consumption and GC execution time cannot have an impact for this kind of micro-benchmarks.

RogerDerekWeiss: I did simple timings using stopwatch statements and it had a 22% speed improvement, so not sure why the performance is not improved on your machine.

```csharp
var a = new int[10_000];
Span<int> span = a;
for (var i = 0; i < 10_000; i++)
{
    for (int j = 0; j < 10_000; j++)
    {
        span[i] = span[i] + j % u;
    }
    span[i] += r;
}
```

> Using 40 as the program argument 
> Execution time without span:  333ms
> Execution time with span: 258ms

run.sh
```diff
@@ -25,3 +25,4 @@ run "PHP" "php ./php/code.php 40"
 run "R" "Rscript ./r/code.R 40"
 run "Python" "python3 ./py/code.py 40" 
 run "Dart" "./dart/code 40"
+run "C#" "./csharp/bin/Release/net9.0/linux-x64/native/code 40"
```
nvasilevski, line 28: Is it possible to configure path to be generic? Otherwise it doesn't seem to be very portable to have version & platform hardcoded in the path

Odonno: Ah yes. I had an idea around that but the best idea I had was to introduce cli args to the bash. And still, I am not sure adding this extra feature would be a good idea because of how the script was intended (i.e. to be executed as it is).

I am open to any idea but IMO it is good enough the way it is.

guibranco: You can do it just by adding the `-o` argument to the  `dotnet publish` command and setting it to a `/bin` path instead of the `.NET version + arch` path.

compile.sh
```diff
@@ -5,3 +5,4 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+dotnet publish ./csharp -c Release -r linux-x64 --self-contained
```
guibranco, line 8:
```suggestion
dotnet publish ./csharp -c Release -r linux-x64 -o ./bin --self-contained
```

guibranco (approved):

run.sh
guibranco, line 28:
```suggestion
run "C#" "./csharp/bin/code 40"
```

guibranco (approved):

run.sh
Odonno: Seems perfect.

Odonno: The only missing part is the target arch, it is still linux-x64. But that should ok for 99% of devs outside Windows I suppose.

compile.sh
```diff
@@ -5,3 +5,4 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+dotnet publish ./csharp -c Release -r linux-x64 -o csharp/bin --self-contained
```
nvasilevski, line 8: https://github.com/bddicken/languages/blob/1c9592da35099d7f1e5cee3c77b7e498fcf19bc8/compile.sh#L8
https://github.com/bddicken/languages/blob/ae5cfceae8d7813291d5dc91d7232a75e0256d53/compile.sh#L8

I'm not familiar with dotnet but folks in other PRs that add C# were able to make the command platform-agnostic

Odonno: Since we run it locally in the same environment, yes, it is indeed possible. Thank you for noticing this.

run.sh
Odonno: Now it is agnostic. Perfect!

valentin-p (approved):

loops/csharp/Program.cs
p6laris: > C and Rust use an array in the stack instead of the heap.
> The size can be determined at the compilation phase so we can use Span in C#.

I’ve always wondered it they allocate the array on the stack, wouldn’t the stack overflow for larger arrays? How do C and Rust handle that?

Odonno: @RogerDerekWeiss Are you running the code in Release mode?

I added a `Stopwatch` and tested with or without Span/stackalloc. I still confirm that I cannot see any difference between the two. It would be interesting to use https://github.com/dotnet/BenchmarkDotNet in order to be sure.

p6laris: > @RogerDerekWeiss Are you running the code in Release mode and with native AOT?
> 
> I added a `Stopwatch` and tested with or without Span/stackalloc. I still confirm that I cannot see any difference between the two. It would be interesting to use https://github.com/dotnet/BenchmarkDotNet in order to be sure.

Did a [benchmark](https://github.com/[p6laris/LangBenchmark](https://github.com/p6laris/LangBenchmark)) with BenchmarkDotNet, It seems the AOT one is slightly faster:

```
BenchmarkDotNet v0.14.1-nightly.20241126.197, Windows 11 (10.0.26100.2454)
Intel Core i7-9750H CPU 2.60GHz (Max: 2.59GHz), 1 CPU, 12 logical and 6 physical cores
.NET SDK 9.0.100
  [Host]     : .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2
  Job-PQNKMO : .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2
  Job-NZKJPK : .NET 9.0.0, X64 NativeAOT AVX2


| Method | Runtime       | Mean    | Error    | StdDev   |
|------- |-------------- |--------:|---------:|---------:|
| Loop   | .NET 9.0      | 1.955 s | 0.0382 s | 0.0571 s |
| Loop   | NativeAOT 9.0 | 1.851 s | 0.0262 s | 0.0232 s |
```

p6laris (approved):

compile.sh
```diff
@@ -5,3 +5,4 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+dotnet publish ./csharp -c Release -o csharp/bin --self-contained
```
neon-sunset, line 8: `PublishSelfContained` is implied by `PublishAot`. Well, it's technically different flavour but you get the idea.

Also, since .NET 8 `publish` defaults to `Release`. We can have just a simple terse `dotnet publish ./csharp -o csharp` nowadays.

bddicken:
Thank you for contributing! However, I decided to accept the C# implementation in #25, so I'm going to close this one out.

PR #23, opened 2024-11-25
[Ruby] Prefer .times over (0...N).each, also add --yjit 

silva96:
Ruby 3.3 and above optimized .times method, also `.times` is the most common way a Well Grounded Rubyist would do this.

Also, I addded --yjit just in case your ruby installation isn't using YJIT. (Edit: it is not enabled by default)

See [this thread ](https://x.com/bsilva96/status/1861136096689606708) for more info

but here's an extract:

> It took 19s with 3.3.6 YJIT (30s w/o YJIT) on my linux-x86_64 machine.
> 
> Changing (0...N).each to N.times made it 9s with 3.3.6 YJIT (8s with 3.4.0-preview2). Using arr.each with arr = (0…N).to_a outside the loops made it 9s with 3.4.0-preview2 YJIT (19s with 3.3.0 YJIT).

> We converted Integer#times to Ruby in 3.3, and Array#each in 3.4. Range#each is still written in C, which YJIT can’t optimize. But we could take a look at it for 3.4. It’s probably Ruby 3.5’s job to make it faster than 8s though.


k0kubun:
> I addded --yjit just in case your ruby installation isn't using YJIT, but I think it does by default in 3.3.6

Ruby 3.3+ with Rails 7.2+ indeed enables YJIT by default, but a regular `ruby` command doesn't just yet. So I appreciate `--yjit` added here.

silva96:
> > I addded --yjit just in case your ruby installation isn't using YJIT, but I think it does by default in 3.3.6
> 
> Ruby 3.3+ with Rails 7.2+ indeed enables YJIT by default, but a regular `ruby` command doesn't just yet. So I appreciate `--yjit` added here.

Thanks @k0kubun , edited the PR description

MarkoZabcic:
This is great. I've tried a simple `while` loop but `times` was faster.

briu:
just FYI It was not enough just to add `--yjit` command in my case

rust should be installed on local machine, without rust ruby returns warning
```
ruby --yjit -v
ruby: warning: Ruby was built without YJIT support. You may need to install rustc to build Ruby with YJIT.
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin21]
```

k0kubun:
`rustc` is not a runtime dependency. You only need `--yjit` as long as you build Ruby with YJIT support. If you install a Ruby binary package with YJIT support, you could get away without installing `rustc`. But yeah, if you build Ruby yourself, you need `rustc` when you build Ruby.

gragtah:
Solid PR, incomplete comparison without these changes imo

jpcamara:
I came to this repo after I noticed a missing --yjit, and I'm so pumped by the work on this PR. And to learn about the huge boosts to .times and .each by writing them in regular ruby and using yjit!

bddicken:
Interesting!

The CEO of Shopify even called me out on not using yjit :)

https://x.com/tobi/status/1861188048542662666

This seems like a good candidate to merge.

PR #24, opened 2024-11-25
feat: add Free Pascal implementation and update build scripts

turric4n:


guaracy:
I think you can try:

 - Compile with `fpc -O3 -XX` to produce smaller and faster executables

-  Use `inline` in function `Fibonacci` like `function Fibonacci(n: Int32): Int32; inline;` to more speed.

guaracy:
Another tip: less comparissions for interation

change 

```pascal
if n = 0 then
  Fibonacci := 0
else if n = 1 then
  Fibonacci := 1
else
  Fibonacci := Fibonacci(n - 1) + Fibonacci(n - 2);
```
by

```pascal
if n < 2 then
  Fibonacci := n
else
  Fibonacci := Fibonacci(n - 1) + Fibonacci(n - 2);
```

bddicken:
This is great, thanks for the contribution.

PR #25, opened 2024-11-25
Add C# support

Noctree:
Added C# code for both loops and fibonacci
Compilation script creates both standard and AOT version.

**Dotnet SDK 9 needed to compile**: https://dotnet.microsoft.com/en-us/download

For compiling the AOT version on Linux/MacOS:
- Make sure you have gcc/clang installed, the sdk needs an external linker to build the executable


loops/csharp/code.cs
```diff
@@ -0,0 +1,10 @@
+﻿var u = int.Parse(args[0]); // Get an input number from the command line
+var r = new Random().Next(10000); // Get a random number 0 <= r < 10k
```
amirvenus, line 2: Use 'Random.Shared' please

amirvenus (changes requested):

loops/csharp/code.cs
```diff
@@ -0,0 +1,10 @@
...
+var a = new int[10000]; // Array of 10k elements initialized to 0
+for (var i = 0; i < 10000; i++) { // 10k outer loop iterations
```
RogerDerekWeiss, line 4: Assign `a` to a new variable of type `Span<int> span` and then mutate the span, its much faster.

fibonacci/csharp/csharp.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
```
TheVeryStarlk, line 6: Please add `<PublishAot>true</PublishAot>` flag.

loops/csharp/csharp.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
```
TheVeryStarlk, line 6: Please add `<PublishAot>true</PublishAot>` flag.

fibonacci/csharp/code.cs
```diff
@@ -0,0 +1,18 @@
+﻿static int Fibonacci(int n) {
+    if (n == 0) {
+        return 0;
+    }
+
+    if (n == 1) {
```
TheVeryStarlk, line 6: Maybe combine the if statement into one?
```cs
if (n <= 1)
    return n;
```

TheVeryStarlk:
Also, please add `<PublishAot>true</PublishAot>` flag to both projects.

TheVeryStarlk:
Nitpick: Follow Allman's bracket style.

fibonacci/csharp/code.cs
```diff
@@ -0,0 +1,18 @@
+﻿static int Fibonacci(int n) {
```
divega, line 1: I know the point isn't to make the code as terse as possible, but I noticed the C++ version appears a bit simpler without unnecessary curly braces, which aren’t needed in C# either. Would you consider expressions-bodied and pattern matching? May seem unfamiliar but have become my preference in recent code and I believe should be as efficient.

``` c#
static int Fibonacci(int n) => n switch
{
    <= 1 => n,
    _ => Fibonacci(n - 1) + Fibonacci(n - 2)
};
```

Noctree:
> Also, please add `<PublishAot>true</PublishAot>` flag to both projects.

AOT is enabled with flags in `compile.sh`, I didn't include it in the project file, so that build types of all the projects are easily togglable in one place

loops/csharp/code.cs
Noctree: Doubt it will make much of a difference performance-wise, Random.Shared still needs to be allocated and initialized when first accessed. And since we're using the random object only once the initialization cost should be the same

fibonacci/csharp/code.cs
Noctree: I don't like using conditional operators in pattern matching, makes the code too confusing for people not familiar with the syntax, IMO.
But a regular switch here is still good, will change

Noctree: Actually, looking at it again, this can be simplified into a ternary operator

fibonacci/csharp/code.cs
Noctree: Good point, done

loops/csharp/code.cs
Noctree: Actually, you were onto something, Random.Shared uses a private Random.ThreadSafeRandom class. I wrote this quick benchmark to check the instantiation performance.

`using System.Linq.Expressions;
using BenchmarkDotNet.Attributes;
using BindingFlags = System.Reflection.BindingFlags;

namespace Benchmarks;

[MemoryDiagnoser]
public class RandomCreationBenchmarks
{
    private static Func<Random> ThreadSafeConstructorCall;

    [GlobalSetup]
    public void Setup()
    {
        var threadSafeRandomClass = typeof(Random).GetNestedType("ThreadSafeRandom", System.Reflection.BindingFlags.NonPublic)!;
        var ctorCall = Expression.New(threadSafeRandomClass.GetConstructor(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance, Type.EmptyTypes));
        ThreadSafeConstructorCall = Expression.Lambda<Func<Random>>(ctorCall).Compile();
    }
    
    [Benchmark]
    public Random CreateParameterlessRandom() => new Random();

    [Benchmark]
    public Random CreateThreadsafeRandom() => ThreadSafeConstructorCall();
}`

Results on my rig are:

image

So Random.Shared does seem to be a lot faster when it comes to instantiation

Noctree: Accidentally ran it using .net SDK 6, here's the correct results

image

Noctree: Accidentally ran the benchmarks with .net 6, here are the correct results

image

tats-u (changes requested):

loops/csharp/code.cs
```diff
@@ -0,0 +1,16 @@
+﻿var u = int.Parse(args[0]);
+var r = Random.Shared.Next(10_000);
+
+//Use span instead of array
+Span<int> a = new int[10_000];
```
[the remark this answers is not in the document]
tats-u: It might have little effect though (allocation occurs just one time)

https://github.com/bddicken/languages/pull/21#discussion_r1859017145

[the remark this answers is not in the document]
neon-sunset: Yup - it's a pointer bump and zeroing in either case (of course goes through an allocation call but it's practically noise). Also there's a trick to avoid the cast or target typing for stackalloc - if you place it in parentheses it will give you a span: `var buf = (stackalloc byte[32]);`.

compile.sh
```diff
@@ -5,3 +5,5 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+dotnet publish csharp/csharp.csproj --ucr --nologo -c Release -o csharp-aot /p:PublishAot=true /p:PublishTrimmed=true
+dotnet publish csharp/csharp.csproj --ucr --nologo -c Release -o csharp
```
neon-sunset, line 9: `publish` now defaults to `Release` automatically so there is no more need to specify it. Same applies to `--use-current-runtime`. Please note that `PublishAot` implies trimming (well, a modified version of it done by ILC). You cannot set it to false, nor need to specify it. both of these can be just `dotnet publish csharp/csharp.csproj -o csharp-aot /p:PublishAot=true` and `dotnet publish csharp/csharp.csproj` respectively.

loops/csharp/code.cs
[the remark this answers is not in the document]
divega: FWIW, I was about to suggest stackalloc too, but I benchmarked it and I confirmed it doesn't make a difference in this case. 

Haven't tried the parenthesis, but I think you get the exact same thing with explicit typing Span<int> instead of var.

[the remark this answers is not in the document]
neon-sunset: There would be no difference, I just didn't know about parentheses until later too and it looks visually annoying to have that one single span to not use `var` among other variables 😅

The execution time here is completely dominated by the loop itself and 1-2s is an eternity compared to ~40KiB array allocation.

loops/csharp/code.cs
```diff
@@ -0,0 +1,16 @@
...
+
+//using a.Length instead of 10_000 constant should give the compiler better optimization hints
+for (var i = 0; i < a.Length; i++)
+{
+    for (var j = 0; j < 100_000; j++)
+    {
+        a[i] = a[i] + j % u;
```
divega, line 12: Nit (because it won't affect perf), but any reason why his isn't using a compound assignment? It could do it in the C++ version as well. Anyway `%` has higher precedence.

```suggestion
        a[i] += j % u;
```

Noctree: Thanks for pointing this out, it is a nitpick, but the code can still look better :)

compile.sh
Noctree: Wasn't sure about the defaults, so I included the arguments just to be sure, thanks for pointing out they're redundant, removed them

TheVeryStarlk:
Is this ready for merge now? 👀🚀

Noctree:
> Is this ready for merge now? 👀🚀

From my side, yes.
I think I addressed all the comments, but it might take a while, considering the amount of new PRs 😅

guibranco (approved):

gldraphael (approved):

TheVeryStarlk:
Little ping, @bddicken. 😅

Noctree:
Merge conflicts resolved

bddicken:
Thanks for this!

loops/csharp/code.cs
```diff
@@ -0,0 +1,16 @@
+﻿var u = int.Parse(args[0]);
+var r = Random.Shared.Next(10_000);
+
+//Use span instead of array
+Span<int> a = new int[10_000];
+
+//using a.Length instead of 10_000 constant should give the compiler better optimization hints
+for (var i = 0; i < a.Length; i++)
```
neon-sunset, line 8: I forgot to mention but using here a.Length over 10_000 constant has no impact since .NET can see that both are identical. It stopped needing most of such hacks over the years :) If you just have an array using a span is also not necessary since arrays is the primary optimization target - array and span optimizations are usually on par with each other not counting some minor differences in codegen.

p6laris: Shouldn't the array stack allocated? Seems like it's faster as mentioned here #21.

neon-sunset: No. There would be no difference - it was discussed in a separate PR too: https://github.com/bddicken/languages/pull/25#discussion_r1859485554

In .NET, `Span<T>` and `ReadOnlySpan<T>` are slice-types much like in Rust. What this means is they do not *own* the memory, they only observe it.

Internally, a span holds `ref T _reference` and `int _length`. That said `ref T` can originate from any source of memory which makes .NET a very unique GC-based platform.

Like other types with contiguous memory, `T[]` has an implicit cast to a `Span<T>`. This means creating a span that holds a byref pointer to the first element of the array and its length. The memory that the span observes still has to live somewhere.

Here, `Span<int> a = new int[10_000];` is a variable declaration and assignment statement. The right-hand side expression in this statement is evaluated first, which performs a heap allocation of an `int[]` array, which is then implicitly converted to a span.

Indeed, there are specially recognized patterns where sometimes array literals immediately consumed by `ReadOnlySpan<T>` which are converted to special readonly data without allocation, but it does not apply here.

To summarize, a span itself can only be placed on stack due to the way GC tracking of `ref T` byref pointers works, but the memory that said span observes can be located anywhere.

p6laris: Thank you so much for the explanation—it's incredibly informative and has helped me resolve some struggles I was having with understanding Span<T> and ReadOnlySpan<T>. :)

I did know that both types merely observe the underlying memory, pointing to the first element of the array. And like when you stackalloc something, the memory is cleared from the stack once the method returns.

However, I hadn’t considered what you pointed out here:

> Like other types with contiguous memory, T[] has an implicit cast to a Span<T>. This means creating a span that holds a byref pointer to the first element of the array and its length. The memory that the span observes still has to live somewhere.

If I understand correctly, when an array is allocated, the GC ensures the array remains alive even after the method’s stack frame is popped. Is that accurate?

This also clarifies why the following doesn’t work:
```csharp
static Span<int> GetSpan()
{
    Span<int> span = stackalloc int[256];
    return span;
}
```
While this does:
```csharp
static Span<int> GetSpan()
{
     Span<int> span = new int[256];
     return span;
}
```

> Indeed, there are specially recognized patterns where sometimes array literals immediately consumed by ReadOnlySpan<T> are then converted to special readonly data without allocation, but it does not apply here.

Yes, your right for the `ReadOnlySpan<T>`, As I’ve heard from Stephen Toub, in those cases, the data gets inlined into the data section of the assembly, which makes sense for readonly literals.

PR #26, opened 2024-11-25
dart: use Uint32List instead of a regular list

ClementBeal:
Hi!

The previous list is quite inefficient when we reassign a value of the "array" quite often or when we try to access it.

`Uint32List` is made for this use case. It's more or less a C array. It has a fixed size and can't be extended. This one only accept unsigned 32-bit integers. We can use 16/32/64 bits if needed.

On my laptop, I have 20% of gain! From ~2500ms to ~2000ms .

### DIFF
--- a/loops/dart/code.dart
+++ b/loops/dart/code.dart
@@ -1,13 +1,12 @@
 import 'dart:math';
+import 'dart:typed_data';
 
 void main(List<String> args) {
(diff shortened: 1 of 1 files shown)


orestesgaolin (approved):

DH-555 (approved):

Feichtmeier (approved):

bddicken:
Nice, ty!

PR #30, opened 2024-11-26
Add Perl

airencracken:
Adds support for benchmarking perl5


loops/perl/code.pl
```diff
@@ -0,0 +1,21 @@
...
+	die "Please input an integer\n";
+}
+
+my $r = int(rand(10000));
+my @a; #let perl allocate the array
+my $i;
+my $j;
+for $i (0..10000) {
+	for $j (0..100000) {
+		$a[$i] = $a[$i] + $j%$n;
```
zmughal, line 15: This can also use `+=`.

loops/perl/code.pl
```diff
@@ -0,0 +1,21 @@
...
+	$n = int(shift @ARGV);
+} else {
+	die "Please input an integer\n";
+}
+
+my $r = int(rand(10000));
+my @a; #let perl allocate the array
+my $i;
+my $j;
+for $i (0..10000) {
```
zmughal, line 13: `10_000-1`

loops/perl/code.pl
```diff
@@ -0,0 +1,21 @@
...
+	for $j (0..100000) {
```
zmughal, line 14: `100_000-1`

loops/perl/code.pl
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/perl
+my $n;
+if(@ARGV) {
+	$n = int(shift @ARGV);
+} else {
+	die "Please input an integer\n";
+}
+
+my $r = int(rand(10000));
+my @a; #let perl allocate the array
```
zmughal, line 10: To match the other implementations:

```perl
my @a = (0) x 10_000;
```

zmughal (changes requested):

loops/perl/code.pl
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/perl
```
zmughal, line 1: Maybe add

```perl
use strict;
use warnings;
```

here too?

airencracken: Didn't see the need. There's a pointless warning for uninitialized variables. I could ignore it specifically, but meh.

loops/perl/code.pl
airencracken: Faster to let perl allocate it dynamically in my tests.

fibonacci/perl/code.pl
```diff
@@ -0,0 +1,24 @@
...
+
+sub factorial {
+	my ($n, $total) = @_;
+
+	$total ||= 1;
+
+	return $total if ($n == 0);
+	@_ = ($n - 1, $n * $total);
+	goto &factorial;
+}
```
bddicken, line 21: This function needs to match the "naive" algorithm that the other implementations use. The idea is not to implement an optimally fast solution, but an equivalent solution.

run.sh
```diff
@@ -25,3 +25,4 @@ run "PHP" "php ./php/code.php 40"
 run "R" "Rscript ./r/code.R 40"
 run "Python" "python3 ./py/code.py 40" 
 run "Dart" "./dart/code 40"
+run "Perl" "./perl/code.pl 40"
```
MuriloChianfa, line 28: can we compile perl before run?

airencracken:
Looks like someone else merged it.

PR #31, opened 2024-11-26
use a function in python

bepvte:
global variables in python are far slower then locals
my testing shows this speeds up `loops` by 2x

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -7,8 +7,11 @@ def fibonacci(n):
     return 1
   return fibonacci(n-1) + fibonacci(n-2)
 
-u = int(sys.argv[1])
-r = 0
-for i in range(u):
-  r += fibonacci(i)
-print(r)
+def main():
+    u = int(sys.argv[1])
+    r = 0
+    for i in range(u):
+      r += fibonacci(i)
+    print(r)
+
+main()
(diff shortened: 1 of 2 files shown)


0xSaltyHash:
yeah testing this on my wsl2 with python3.12 gave me 45.370 S
image

james598keen:
Hello,
You're right! Local variables in Python are indeed faster than global variables because local variable access involves fewer lookup steps. This optimization can significantly speed up loops, potentially doubling the performance as you've noticed.

Best Regards,
James Keen

fibonacci/py/code.py
```diff
@@ -7,8 +7,11 @@ def fibonacci(n):
...
   return fibonacci(n-1) + fibonacci(n-2)
 
-u = int(sys.argv[1])
-r = 0
-for i in range(u):
-  r += fibonacci(i)
-print(r)
+def main():
+    u = int(sys.argv[1])
+    r = 0
```
oriontvv, line 12: Pythonic way here is to use builtin `sum` function. And type hints would be nice to have.

bddicken: The idea is to implement the "same" way across all implementations, so will not be adding sum at this time.

bddicken:
Wow cool, thanks for the contribution.

PR #33, opened 2024-11-26
Assembly

trinidadgithub:


bddicken:
It's really cool that you created an asm version of this. However, I'd like these benchmarks to be portable, where others could run it easily on various operating systems and architectures. This would not be feasible with adding a direct asm implementation.

Again, super cool contribution, but I don't think its realistic for me to merge this in.

trinidadgithub:
Closing PR for now.

PR #34, opened 2024-11-26
Add cpp loop version

hare1039:
Pass by and add the std modern c++ version.

### DIFF
--- a/loops/cpp/code.cpp
+++ b/loops/cpp/code.cpp
@@ -1,0 +1,24 @@
+#include <string>
+#include <random>
+#include <array>
+#include <print>
+#include <cstdint>
+#include <ranges>
+
+int main (int argc, char* argv[])
+{
+    int const mod = std::stoi(argv[1]);                         // Get an input number from the command line
+    std::random_device rd{};
+    std::mt19937 engine{rd()};
+    std::uniform_int_distribution<int> dist{0, 9999};           // <random> library setup
+
+    int const picked_number = dist(engine);                     // Get a random number 0 <= r < 10k
+    std::array<std::int32_t, 10000> array{};                    // Array of 10k elements initialized to 0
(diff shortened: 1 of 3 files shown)


loops/cpp/code.cpp
Hxnter999, line 1: looks good; consider `std::println` though.

loops/cpp/code.cpp
```diff
@@ -0,0 +1,24 @@
...
+    int const mod = std::stoi(argv[1]);                       // Get an input number from the command line
+    std::random_device rd{};
+    std::mt19937 engine{rd()};
+    std::uniform_int_distribution<int> dist{0, 9999};         // <random> library setup
+
+    int const picked_number = dist(engine);                   // Get a random number 0 <= r < 10k
+    std::array<std::int32_t, 10000> array{};                  // Array of 10k elements initialized to 0
+    for (int const i : std::ranges::iota_view{0, 10000})      // 10k outer loop iterations
+    {
+        for (int const j : std::ranges::iota_view{0, 100000}) // 100k inner loop iterations, per outer loop iteration
```
bddicken, line 19: Idk what iota_view is, can you explain?

safetycoomer: a range type, iterable over the provided values https://en.cppreference.com/w/cpp/ranges/iota_view

loops/cpp/code.cpp
hare1039: Thanks for the suggestion. Updated the code.

loops/cpp/code.cpp
hare1039: Its similar to python's `range(0, 10000)`

hare1039: Switched to `std::ranges::views::iota` for clarity.

bddicken:
Thanks for contributing!

PR #36, opened 2024-11-26
Add Elixir

klirix:
This is a recursion based Elixir implementation. Initially implemented as streams, it was slow, then using reduce, then came to a solution based on recursive function calls. Implementation using for loops doesn't really make sense since the data in elixir is immutable

fibonacci:
```
time elixir elixir/bench.exs 40
165580140

real    0m1.982s
user    0m2.660s
sys     0m0.305s
```
loops:
```
time elixir elixir/bench.exs 40
1952939

real    0m3.490s
user    0m4.178s
sys     0m0.251s
```


michelson:
https://gist.github.com/michelson/fcf05b0f7a7fe1aba95b98a5efff43ee#file-bench-ex
This elixir version takes an Execution Time: of 1.025 seconds

fibonacci/elixir/bench.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  def sum_fibs(n), do: sum_range(0, 0, n)
+
+  defp sum_range(0, 0, n), do: sum_range(1, 0, n)
+
+  defp sum_range(i, acc, n) when i < n do
+    sum_range(i + 1, acc + fib(i), n)
+  end
+
+  defp sum_range(_, acc, _), do: acc
```
DeedleFake, line 22: According to some quick testing with `:benchee`, this version's about 0.118s faster on my machine.

```suggestion
  def run do
    cli_num = System.argv() |> Enum.at(0) |> String.to_integer()
    fib_res = sum_fibs(cli_num)
    IO.puts(fib_res)
  end

  defp sum_fibs(i \\ 0, max, total \\ 0)
  defp sum_fibs(i, max, total) when i < max, do: sum_fibs(i + 1, max, total + fib(i))
  defp sum_fibs(_i, _max, total), do: total

  defp fib(0), do: 0
  defp fib(1), do: 1
  defp fib(n), do: fib(n - 1) + fib(n - 2)
```

loops/elixir/bench.ex
DeedleFake, line 1: I think that there is quite a bit of room for optimization in here. I'll take a look at it when I get a chance to.

compile.sh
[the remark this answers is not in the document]
klirix: I actually got no idea how to implement this without awkward directory switching, since iex\elixir executables load the beam modules in the current dir. I tried to find the solution to this, but wasn't able to :(

fibonacci/elixir/bench.ex
klirix: Your version is way cleaner!
I haven't found any significant improvements on my machine though. It looks like it cuts one function call after all 🤔

loops/elixir/bench.ex
MikaelFangel: See: https://github.com/bddicken/languages/pull/36#discussion_r1858536055

compile.sh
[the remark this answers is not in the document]
DeedleFake: I haven't tested this, but you should be able to use the `-pa` flag, i.e. `-pa elixir`.

loops/elixir/bench.ex
```diff
@@ -0,0 +1,54 @@
...
+  # 10k outer loop iterations
+  defp outer_loop_body(i, acc, random_num, cli_num) when i < 10_000 do
+    # Add a random value to each element in array
+    val = random_num + inner_loop(cli_num)
+    outer_loop_body(i + 1, Tuple.insert_at(acc, i, val), random_num, cli_num)
+  end
+
+  defp outer_loop_body(10_000, acc, _, _) do
+    acc
+  end
```
DeedleFake, line 53: Here's what I've got so far. To my utter shock, this is benchmarking ever so slightly _slower_ on my machine and I'm not sure why. It does, however, fix a bug in the original code where it should have been using `put_elem/3` instead of `Tuple.insert_at/3`. That issue is moot in this version as it just builds a list as it goes instead of modifying elements of a tuple.

```suggestion
  def run() do
    [input | _] = System.argv()
    input = input |> String.to_integer()
    rand = :rand.uniform(10_000)
    result = outer(rand, input)
    IO.puts(result)
  end

  defp outer(rand, input) do
    for _ <- 0..9_999//1 do
      inner(input) + rand
    end
    |> Enum.at(rand)
  end

  defp inner(i2 \\ 0, input, result \\ 0)
  defp inner(100_000, _input, result), do: result

  defp inner(i2, input, result) do
    result = result + rem(i2, input)
    inner(i2 + 1, input, result)
  end
```

josevalim:
Nice work @klirix, some notes:

1. You don't need to compile the Elixir file. You can just run the script directly. Here is what I wrote for fib:

```
defmodule Math do
  def fib(0), do: 0
  def fib(1), do: 1
  def fib(n), do: fib(n - 1) + fib(n - 2)

  def loop(0, res), do: IO.puts(res)
  def loop(i, res), do: loop(i - 1, res + fib(i))
end

[num] = System.argv()
Math.loop(String.to_integer(num) - 1, 0)
```

2. We should not use tuples for loops in Elixir. Because the initial benchmark assumes mutability, it is unclear if that's a requirement, so we have two options:

a. Build a list as we build the loops
b. Use the `:counters` module (https://www.erlang.org/doc/apps/erts/counters.html), which is literally a mutable array

Perhaps option `b` will keep the spirit of the benchmark.

klirix:
Thanks @josevalim
> You don't need to compile the Elixir file. You can just run the script directly
Wouldn't precompilation save time on initialization times and allow for optimizations?

> 2. We should not use tuples for loops in Elixir. Because the initial benchmark assumes mutability, it is unclear if that's a requirement, so we have two options:
I somehow got slightly better times using the Tuples instead of the Lists in the interation itself, but also I used it for faster random element access in the end of the benchmark itself. My initial implementation was using Lists but I thought that preallocation would make sense here, even though it goes against idiomatic solutions

> b. Use the `:counters` module (https://www.erlang.org/doc/apps/erts/counters.html), which is literally a mutable array
> Perhaps option `b` will keep the spirit of the benchmark.
I was thinking the same about the spirit of menchmark maybe using counters module would be a better solution

josevalim:
Scripts are also compiled before running, so runtime performance is the same. And the file is small enough that the time to compile it is negligible.

counters should be more efficient than tuples too :)

DeedleFake:
@josevalim

a. I tried doing that and it wound up _slower_ than the tuple-based version. I've been trying to figure out why. See this comment: https://github.com/bddicken/languages/pull/36#discussion_r1858817151

b. Ooh. I tried using `:ets.update_counter/4` and that _kind of_ worked, but it was both kind of ugly code and quite slow. `:counters` is good to know about. Unfortunately, I tried building a version that used that it was a lot slower than just building a list:

```elixir
defmodule New do
  def run() do
    [input | _] = System.argv()
    input = input |> String.to_integer()
    rand = :rand.uniform(10_000)
    result = outer(rand, input)
    IO.puts(result)
  end

  defp outer(i \\ 1, rand, input),
    do: outer(i, rand, input, :counters.new(10_000, [:write_concurrency]))

  defp outer(10_001, rand, _input, counters), do: :counters.get(counters, rand + 1)

  defp outer(i, rand, input, counters) do
    :counters.add(counters, i, inner(input) + rand)
    outer(i + 1, rand, input, counters)
  end

  defp inner(i2 \\ 0, input, result \\ 0)
  defp inner(100_000, _input, result), do: result

  defp inner(i2, input, result) do
    result = result + rem(i2, input)
    inner(i2 + 1, input, result)
  end
end
```

Benchee results:

```
Benchmarking new ...
1954561
1950948
1955270
1957398
Benchmarking old ...
1950065
1950118
1955649
1957186
Calculating statistics...
Formatting results...

Name           ips        average  deviation         median         99th %
old           0.49         2.03 s     ±0.10%         2.03 s         2.03 s
new           0.42         2.41 s     ±0.81%         2.40 s         2.43 s

Comparison:
old           0.49
new           0.42 - 1.19x slower +0.38 s
```

josevalim:
@DeedleFake the list one may be slower because of the range? You can try this:

```
@list Enum.to_list(1..9999)

def list(...) do
  for _ <- @list do

  end
end
```

There is a chance the cost is in the protocol, which is not consolidated if you are not inside the project.

klirix:
@michelson 

My intuition coming from other languages was telling me that dynamic allocations could be a culprit here. I know there are tons of optimizations for tuples in the vm for inserting and fetching the tuple elements and it's not actually allocating the tuple 10k times. But the list is actually allocating each node on iteration, so that makes for 10k miniscule allocation operations.

In all fairness 10k is a pretty small scale anyway to see the effects

fibonacci/elixir/bench.ex
DeedleFake: @MikaelFangel

It's fairly easy to do it iteratively instead of recursively, yes, but the benchmark is supposed to be done recursively. Take a look at some of the other languages' implementations, such as Go. Those are standard imperative languages with mutable state and yet they do it recursively, too.

DeedleFake:
@klirix

You should actually be using `put_elem/3`, not `Tuple.insert_at/3`. You have an existing tuple with 10,000 zeros. It should be changing them, not inserting new ones.

DeedleFake:
It seems that I've misunderstood the way that pattern matching and guard clauses are optimized. It turns out that this order is quite a bit faster, making it the fastest one that I've tested so far:

```elixir
  def run() do
    [input] = System.argv()
    input = input |> String.to_integer()
    rand = :rand.uniform(9_999) - 1
    result = outer(rand, input)
    IO.puts(result)
  end

  defp outer(i \\ 0, rand, input, results \\ [])

  defp outer(i, rand, input, results) when i < 10_000 do
    result = inner(input) + rand
    outer(i + 1, rand, input, [result | results])
  end

  defp outer(_i, rand, _input, results), do: results |> Enum.at(10_000 - rand - 1)

  defp inner(i2 \\ 0, input, result \\ 0)

  defp inner(i2, input, result) when i2 < 100_000 do
    result = result + rem(i2, input)
    inner(i2 + 1, input, result)
  end

  defp inner(_i2, _input, result), do: result
```

fibonacci/elixir/bench.ex
```diff
@@ -0,0 +1,15 @@
...
+    IO.puts(fib_res)
+  end
+
+  defp sum_fibs(i \\ 0, max, total \\ 0)
+  defp sum_fibs(i, max, total) when i < max, do: sum_fibs(i + 1, max, total + fib(i))
+  defp sum_fibs(_i, _max, total), do: total
+
+  defp fib(0), do: 0
+  defp fib(1), do: 1
+  defp fib(n), do: fib(n - 1) + fib(n - 2)
```
DeedleFake, line 14: Managed to get it down a bit more using what I found about pattern matching vs. guard clauses that I mentioned in another comment:

```suggestion
  defp fib(n) when n > 1, do: fib(n - 1) + fib(n - 2)
  defp fib(n), do: n
```

fibonacci/elixir/bench.ex
MikaelFangel: I know it’s supposed to be done recursively, hence my suggestion. Just to clarify, tail call optimizations (TCO) are just a specific form of recursion and not an iterative solution.

The other solutions mentioned can also be written in a uniform way that allows for TCO, which aligns with the rationale behind PR #55. This approach would better simulate real-world scenarios, where efficient recursion is often critical. A uniform TCO-enabled implementation across languages would make the benchmark both fairer and more reflective of practical usage.

fibonacci/elixir/bench.ex
klirix: Oh wow the guards are noticably more performant 🤔

fibonacci/elixir/bench.ex
klirix: @MikaelFangel This is amazing! It makes the fibonacci down all the way from 1.8s to 400ms on my machine

I don't see why we shouldn't include it

DeedleFake: The solution you gave _is_ an iterative Fibonacci solution. The fact that it is written using recursive function calls is irrelevant. It is equivalent to doing

```go
a, b := 0, 1
for range n {
  a, b = b, a + b
}
return a
```

in Go. The entire point of the test is to cause the exponential recursive growth. If you do it the way that you're suggesting, it's not the same benchmark anymore.

compile.sh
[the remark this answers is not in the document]
klirix: I opted instead to run exs directly, instead of compilation approach

fibonacci/elixir/bench.exs
```diff
@@ -9,8 +9,9 @@ defmodule Bench do
   defp sum_fibs(i, max, total) when i == max, do: total
   defp sum_fibs(i, max, total), do: sum_fibs(i + 1, max, total + fib(i))
 
-  defp fib(n) when n <= 1, do: n
-  defp fib(n), do: fib(n - 1) + fib(n - 2)
+  defp fib(n, current \\ 0, next \\ 1)
+  defp fib(n, current, _next) when n == 0, do: current
+  defp fib(n, current, next), do: fib(n - 1, next, current + next)
```
DeedleFake, line 14: This is an iterative solution. No other implementations in this repo use an iterative solution. You'd have to ask @bddicken to be sure, but I'm pretty darn sure that the intention of the benchmark is to do it the inefficient, recursive way.

klirix: I would agree that we need the opinion of the repo maintainer for this

But so far I don't see the problem, as per definition, fib benchmark case `Emphasizes function call overhead and recursion`, which is exactly what we're doing here

DeedleFake: No it isn't. The entire point of tail-call optimization is that it _doesn't_ have function call overhead. The compiler optimizes away the recursion and turns it into a loop. There is no recursion happening there after compilation.

bddicken: Thanks for your contribution!

We should only include solutions that work the "same" as the others. If we allowed this, then we'd have to open up all the other version for further optimization with tail recursion and/or iterative solutions.

Thus, as it stands now, this is disqualified. If you want to implement it in a way that matches the other solutions, send me a follow up.

klirix: Alright, makes sense, I'll put back the other solutions back shortly

bddicken:
Cool, thanks for submitting this.

PR #38, opened 2024-11-26
add jitless node, deno and bun

twlite:
This pull request includes changes to the `run.sh` script to add support for running certain commands in a "jitless" mode and to pass environment variables to the `runOnce` function.

Enhancements to `run.sh`:

* Modified the `runOnce` function to accept an additional parameter for environment variables, allowing commands to be run with specific environment settings.
* Added support for running Node, Bun, and Deno commands in a "jitless" mode by adding new entries for these modes in the `run` function.


twlite:
added bun jitless as well, this one needed modifications to run and runOnce to accept third argument for env

bddicken:
Nice thanks!

PR #41, opened 2024-11-26
Add Nim, fix run.sh

xTrayambak:
- **(add) add Nim benchmark**
- **(add) add Nim to compile.sh**
- **(fix) use `/usr/bin/env time`, not `/usr/bin/time`**

`run.sh` wouldn't run on NixOS because it was expecting `/usr/bin/time`, which is not going to be present on non-FHS distros. Now, it uses `/usr/bin/env` to locate the binary instead.


bddicken:
Thanks for contributing. Would you mind adding a solution for both of the existing benchmarks?

xTrayambak:
Sure.

xTrayambak:
@bddicken, @Beetroit's solution for fibonacci looks fine. I only have a problem with his loops implementation because he's explicitly allocating a vector to the heap, which is a big no-no unless you really have to.

xTrayambak:
@bddicken Sorry if I'm being pushy, could you merge this? The fibonacci implementation in Nim is fine, but my loop "implementation" is better.

xTrayambak:
Also, @Beetroit is there a particular reason you are building the Nim program in debug mode? Just adding `-opt:speed` doesn't trigger all the optimizations as far as I know..

Beetroit:
I belive the inner loop should be 100k items, not 10k.

Beetroit:
what's your time? For the loop.
My implementation averages around 3.5
image

xTrayambak:
image
This is on a Ryzen 5 5500H, so not that powerful.

xTrayambak:
Again, the stack will always be faster than the heap, so `newSeq` is really not a good solution unless you're dealing with something like a huuuuge vector that'd overflow the stack's limits. You generally don't reach that limit easily unless you're doing huge amounts of data processing.

Beetroit:
I modified the code a bit to use arrays. Run my new pr on your end let's see. I'm getting 6+ secs from your code.

BostonBoston:
I was going to make a PR but I'll just add that when I need to optimize I use ` -d:danger -d:lto --passC:"-march=native"` 
`--mm:none` is a case by case basis but without a seq I think the only thing that would use GC'd mem is the parseInt call

bddicken:
I'm not sure if this is needed anymore. Seems like the main change is the switch to `+=`. Does that make a bug difference performance wise here?

bddicken:
Looks like there's been other improvements to nim implementation since this, so I am going to close it out. Thank you for submitting this though.

PR #42, opened 2024-11-26
Make Python More Gooder

thedanvail:
Use list comprehension to make python betterer than before


thedanvail:
Closed in favor of https://github.com/bddicken/languages/pull/53

PR #46, opened 2024-11-26
Add Lua

botantony:
Lua benchmarks


juliamertz:
For LuaJIT it would probably make more sense to turn in into bytecode first with `luajit -b lua/code.lua lua/code`

L1Q:
Also for luajit, consider using ``require("table.new")`` and ``local a = table.new(0, 10000)``.

SerJaimeLannister:
Hey interesting that I had written the same thing except my code was strictly inspired by the golang code base and I am new to lua , so maybe you should check if mine has some bottleneck since your code seems ways simpler

Also another thing I wanted to ask was , if you run something like 
hyperfine --warmup 3 \
          --min-runs 5 \
          --show-output \
          "lua main_optimized.lua 40"

then since os.time() doesn't get updated as fast , I think this can cause the same result twice or maybe even more times so I am not sure where my implementation has tried to fix it.

bddicken:
Nice, ty!

PEZ:
Would be sweet with Lua ported to the new in-process runner!

* #371

PR #47, opened 2024-11-26
Link to https://benjdd.com/languages/

vdm:
Also should be set in github repository [About Website](https://stackoverflow.com/a/62767178/392033) field.

An embed in README.md would be even better.


bddicken:
TY.

PR #48, opened 2024-11-26
Add SIMD C

huppfi:
The C Code used in the Repo is very unoptimized. Using SIMD we should be able speed up the code by a lot.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,4 +1,5 @@
 clang -O3 c/code.c -o c/code
+clang -mavx2 -O3 -mfma cfast/code.c -o cfast/code
 go build -o go/code go/code.go
 javac jvm/code.java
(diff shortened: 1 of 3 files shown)


bddicken:
This is super cool, but also I know almost nothing about SIMD. Would you mind adding some explanation / context for the code?

enghitalo:
This will be nice to see!

axman6:
This really doesn't feel like it's at all within the spirit of this project, it doesn't demonstrate "C", it demonstrates x86_64 assembly using C as a macro language. If you're going to go to these lengths, then why not just compute the answer directly using the relatively simple equation for `r + sum([0..10000] mod r)`? 
Also, it looks like this could be more efficient if you used Lemire's algorithm for fast modulus instead of the old multiply+shift algorithm.

bddicken:
I do think this is quite cool, but for now I'm considering simd "out of scope" for this, so going to close out.

PR #53, opened 2024-11-26
10x Python perf with `sum` and `itertools`

soof-golan:
Just to make Python less shamefully slow

## Changes

* Sum up the `i % u` range as an iterator. This yields about 10-20x improvement. (On my M1 mac: ~4.2 seconds)
* Introduce a function scope to reduce variable lookup time (minor)

And if we're allowed to "cheat" by optimizing by hand, then this reduces Python to about 10-20ms:

```python
import sys
from itertools import cycle, islice, repeat
from random import randint


def main():
    u = int(sys.argv[1])  # Get an input number from the command line
    r = randint(0, 10000)  # Get a random number 0 <= r < 10k
    # 10k outer loop iterations
    a = list(repeat(sum(islice(cycle(range(u)), 100000)) + r, 10000))
    print(a[r])  # Print out a single element from the array

if __name__ == "__main__":
    main()
```

### DIFF
--- a/loops/py/code.py
+++ b/loops/py/code.py
@@ -1,10 +1,21 @@
 import sys
-import random
-u = int(sys.argv[1])         # Get an input number from the command line
-r = random.randint(0, 10000) # Get a random number 0 <= r < 10k
-a  = [0] * 10000             # Array of 10k elements initialized to 0
-for i in range(10000):       # 10k outer loop iterations
-  for j in range(100000):    # 100k inner loop iterations, per outer loop iteration
-    a[i] = a[i] + j%u        # Simple sum
-  a[i] = a[i] + r            # Add a random value to each element in array
-print(a[r])                  # Print out a single element from the arra
+from itertools import cycle, islice
+from random import randint
+
+
+def main():
+    u = int(sys.argv[1])  # Get an input number from the command line
+    r = randint(0, 10000)  # Get a random number 0 <= r < 10k
+    a = [0] * 10000  # Array of 10k elements initialized to 0
+    # 10k outer loop iterations
+    for i in range(len(a)):
+        # 100k inner loop iterations, per outer loop iteration
+        a[i] = sum(islice(cycle(range(u)), 100000))
+        # Add a random value to each element in array
+        a[i] += r
+
+    print(a[r])  # Print out a single element from the array
+
+
+if __name__ == "__main__":
+    main()


juestr:
The outer "loop" here just repeats the *result* from the inner one 100k times, this isn't a fair optimization but a rewrite possible in any language.  It also leaves out the % operation.

I don't think you understand what repeat does, the inner islice and cycle are no-ops here.

george-gca:
> this isn't a fair optimization but a rewrite possible in any language

I agree in the sense that this is a optimized version of the specific code presented in the example, not of "any regular loop".


Now, since the `i` variable in the original code is only used to access the list position, this code creates "equivalent" result to the original one (given the randomness). This can be easily validated by sharing the values of `u` and `r` and doing this:

```python
from itertools import cycle, islice, repeat
from random import randint

u = 40                        # Get an input number from the command line
r = randint(0, 10_000)        # Get a random number 0 <= r < 10k

# original code
a  = [0] * 10_000             # Array of 10k elements initialized to 0
for i in range(10_000):       # 10k outer loop iterations
  for j in range(100_000):    # 100k inner loop iterations, per outer loop iteration
    a[i] = a[i] + j%u         # Simple sum
  a[i] = a[i] + r             # Add a random value to each element in array
print(a[r])                   # Print out a single element from the array

# this code
b = list(repeat(sum(islice(cycle(range(u)), 100_000)) + r, 10_000))
print(b[r])

print(a == b)
```

which outputs (in my case with `u = 40` and `r = 8139`):

```
1958139
1958139
True
```

george-gca:
But note that the code in the PR is different from the one pasted above, so I believe the one in the PR is a fair comparison, if the `%u` is added.

bddicken:
The idea of this benchmark is not to implement the fastest possible version of the code to "compute the same thing". If that were the case, we could make all of the other implementations faster as well. The idea is to write equivalent (or very close to equivalent) implementations of the same thing and then compare runtimes.

As it is currently written, I can't accept this PR (even if the current version is not very "pythonic").

george-gca:
I understand and agree. What about the implementation in #42? I believe it is closely related to the original code.

juestr:
The result of the following would also return the same result:

    u = 40
    r = 8139  # randint(0, 10_000)
    v = 0
    for j in range(100_000):
        v = v + j%u
    v += r 
    a = [v for _ in range(10_000)]
    print(a[r]) 

The so-called inner loop on j is NOT executed in a nested fashion but only ONCE up front. This works because yes, all elements of a (and b) are equal. And the code in the PR exhibits the exactly same problem:

sum() is only called ONCE, and its RESULT VALUE is repeated, NOT the whole execution of the sum()-expression. That's how function argument evaluation works and always has worked.

I do say I blame the original benchmark code for being too simple, making each value of a[] at least dependent on i and j wouldn't have been hard and would easily have avoided this problem.

soof-golan:
> As it is currently written, I can't accept this PR (even if the current version is not very "pythonic").

That's understandable and quite unfortunate as pure-Python iteration is quite slow (as we can all see).

PR #55, opened 2024-11-26
Made all fibbonaci test tail recursive

PhilipCramer:
Since the way the Fibonacci solutions were written caused exponential recursion tree growth.
All language solutions were rewritten to a tail call solution.
This evens out the test to focus on efficient recursion handling rather that testing stack sizes.
By doing this the test can clearly show if the language supports efficient recursions.

### DIFF
--- a/fibonacci/c/code.c
+++ b/fibonacci/c/code.c
@@ -2,10 +2,12 @@
 #include "stdlib.h"
 #include "stdint.h"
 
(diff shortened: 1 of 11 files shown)


bddicken:
Thanks for the idea here, but I am going to keep the implementations as they are. It is a naive solution, but in this case it's actually nice to test a significant amount of function calls and stack sizes.

PR #56, opened 2024-11-26
add julia

ninjaaron:
I added a benchmark for Julia :)


ninjaaron:
Ah, I see there is already one for Julia!

PEZ:

* #371
* #410

PR #57, opened 2024-11-26
Use int32 instead of int

CalebQ42:
Using int32 instead of int offers a significant performance uplift in Go. On my computer it went from 1.51s to 1.29s (tied with C/Rust)


CalebQ42:
Duplicate of #44

PR #58, opened 2024-11-26
Updated code.go to use go's concurrency features

Trust-genie:
Hi there, 
I observed that Go could do much better than what you have right now. Go's main strength is its concurrency models. The ability to take advantage of multiple cpu's when performing an operation. For such a (heavy) computation, writing single threaded go is a mistake. I do understand that the project is just to test and so maybe writing the code single threaded is the idea, if this is the case then I apologise in advance. I also have a bench file comparing the two approaches.(on second thought  i should already have included it with the pull request)
Anyways thank you for you for all you do.

Love Liezner

### DIFF
--- a/loops/go/code.go
+++ b/loops/go/code.go
@@ -1,22 +1,86 @@
+// Hello there
+//
+// Go's main strenght is its concurrency model and writing go without that in mind is IMO a mistake
+// i made some considerations since just spawning go routines would be careless,
+// So I will be performing operation in batches
+//
+// Also int and uint types behave slightly differently in go. Here I choose to use uint because
+// 1. Our values are guaranteed to be positive and so using uint results in more size and less chance of an overflow
+// 2. by using uint (and not uint32 or uint64) you ensure that at compile time, values defualt to system arch size of say 32 or 64 bits
+// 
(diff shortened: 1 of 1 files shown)


mglsj:
> Hi there, I observed that Go could do much better than what you have right now. Go's main strength is its concurrency models. The ability to take advantage of multiple cpu's when performing an operation. For such a (heavy) computation, writing single threaded go is a mistake. I do understand that the project is just to test and so maybe writing the code single threaded is the idea, if this is the case then I apologise in advance. I also have a bench file comparing the two approaches.(on second thought i should already have included it with the pull request) Anyways thank you for you for all you do.
> 
> Love Liezner

I believe the test was designed to show single threaded performance of each language. For a fair comparison concurrency would have to be implemented for all languages.

Trust-genie:
If that's the case then we could still do better.
I'll make another pull request

bddicken:
Thanks for the contribution.

 If I allowed for these types of optimizations (multi-threading / multiprocess), then to be fair, I'd have to open up all the other implementations to such improvement as well. For now I'm not going to do that, so will be closing this out.

PR #65, opened 2024-11-26
Speed up javascript with eval

serpent7776:
This change speeds up the javascript loops when run in bun on my machine.

Before the change

```
Node = 2.41users
0inputs+0outputss
Node = 2.39users
0inputs+0outputss
Node = 2.42users
0inputs+0outputss

Bun = 2.09users
0inputs+0outputss
Bun = 2.10users
0inputs+0outputss
Bun = 2.08users
0inputs+0outputss
```

After the change

```
Node = 2.39users
0inputs+0outputss
Node = 2.42users
0inputs+0outputss
Node = 2.39users
0inputs+0outputss

Bun = 1.84users
0inputs+0outputss
Bun = 1.84users
0inputs+0outputss
Bun = 1.87users
0inputs+0outputss
```

### DIFF
--- a/loops/js/code.js
+++ b/loops/js/code.js
@@ -1,10 +1,11 @@
 const u = Number(process.argv[2]);           // Get an input number from the command line
 const r = Math.floor(Math.random() * 10000); // Get a random number 0 <= r < 10k
 const a = new Int32Array(10000);             // Array of 10k elements initialized to 0
+eval(`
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for your contribution. It's quite interesting that this actual speeds things up for Bun. However, this is more of a hack than it is an actual improvement based on language features, so I will not be incorporating this into the main benchmark at this time.

PR #66, opened 2024-11-26
Add Clojure benchmarks

NoahTheDuke:


MTrost:
@NoahTheDuke 

I think the idea of this repository was comparing performance of similar work being done in different languages, i.e. doing the same steps as the C example code, even if it's not how we normally write idiomatic Clojure.

Also your inner "loop" only does 10k instead of 100k elements.

NoahTheDuke:
It's a poor benchmark if the code isn't idiomatic. If that's the intended case tho, then we can close this because a version with `loop` and `int-array` and `aset` will generate nearly 1-to-1 code as the existing Java version.

PR #67, opened 2024-11-26
Fix the unfair betwenn C and Rust code (unsigned 32 bit interger) and use stable toolchain

Paulo-21:
Hello,
I have fixed the rust performance problem, the original code was using size  as the size of integer for the tab A and the C code is using integer of 32 bits.
The size type is the type that matches the size of a pointer to refer to any location in memory, so 32 bits on 32 bit platform and 64 bits on most computer today.

I have fixed  that by just replacing usize by u32
So compiler llvm for both can optimize the code the same way.

On my machine
Before: 
C: 1.83 sec
Rust: 2.01 sec
After: 
C: 1.83 sec
Rust: 1.70 sec

Cool!

I have also removed the necessary of nightly build so every one can test and removed the-Zlocation-detail=none flag which is only supported by nightly and does nothing for performance for this type of programme and it's generally nano benefits.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,7 +1,7 @@
 clang -O3 c/code.c -o c/code
 go build -o go/code go/code.go
 javac jvm/code.java
-RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Cargo.toml --release
+cargo build --manifest-path rust/Cargo.toml --release
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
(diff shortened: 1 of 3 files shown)


Paulo-21:
The second optimisation give me 
Before : 
C : 0,36 sec
Rust : 0,47 sec
After : 
Rust : 0,36 sec (same as C)

bddicken:
Thanks! I was trying to figure out how to remove the nightly configuration previously but could not figure it out (I'm not a rust guy). This is great.

PR #69, opened 2024-11-26
PR validation workflow added

rsb-23:
**Idea** behind this PR is to automate _benchmarking the code changes_ and _speed-up PR merges_.

**Workflow** first lists modifed files, then for each file it logs execution time using `timeit.sh` and all benchmark results are added as PR comment.

### Notes :
- No existing file is modified. (timeit.sh is just copy of runOnce function)
- This PR is incomplete, hence merge to new-branch say `pr-benchmark`, not `main`.

---
If this looks nice, we can complete it.

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+file=$1
+
+function timeit {
+  bash timeit.sh "$1" "$2 40"  >> runtime.txt
+}
+
+if [[ $file == */js/* ]]; then
+  echo "Testing JavaScript file: $file"
+  timeit "Node" "node $file"
+  timeit "Bun" "bun $file"
+  timeit "Deno" "deno $file"
+elif [[ $file == */php/* ]]; then
+  timeit "PHP" "php $file"
+elif [[ $file == */py/* ]]; then
+  echo "Testing Python file: $file"
+  timeit "PyPy" "pypy $file"
+  timeit "Python" "python3 $file"
+elif [[ $file == */r/* ]]; then
+  timeit "R" "Rscript $file"
+elif [[ $file == */ruby/* ]]; then
+  timeit "Ruby" "ruby $file"
+else
+  echo "No tests defined for $file"
+fi
--- a/.github/workflow/benchmark.yml
+++ b/.github/workflow/benchmark.yml
@@ -1,0 +1,56 @@
+name: Test Modified Files on PR
+
+on:
+  pull_request:
+
+jobs:
+  test-modified-files:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Step 1: Check out the code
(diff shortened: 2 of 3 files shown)


run-test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+file=$1
+
+function timeit {
```
EmmmaTech, line 4: why is the actual timing function defined in another shell script? it seems unnecessary when no other scripts depend on that script.

run-test.sh
```diff
@@ -0,0 +1,25 @@
...
+elif [[ $file == */php/* ]]; then
+  timeit "PHP" "php $file"
+elif [[ $file == */py/* ]]; then
+  echo "Testing Python file: $file"
+  timeit "PyPy" "pypy $file"
+  timeit "Python" "python3 $file"
+elif [[ $file == */r/* ]]; then
+  timeit "R" "Rscript $file"
+elif [[ $file == */ruby/* ]]; then
+  timeit "Ruby" "ruby $file"
```
EmmmaTech, line 22: this feels like a burden on the developers to have to update 2 files in order to add a new language. an ideal solution is for `run.sh` to accept an optional argument where the language to test is specified, where the logic here can kick in.

rsb-23:
@EmmmaTech 
Thanks for your review. 
But as I mentioned, this is an incomplete PR or say just an idea. It implies there are redundancy and errors.

This PR is for maintainer(s) to try it out and see if it fits in the project.
If it deemed suitable to maintainers, I can move ahead.

EmmmaTech:
I see, I didn't think it would imply that. would marking the pr as a draft make sense for that tho?

rsb-23:
Actually, I need reviews/discussion on the **idea** hence its marked as PR. (code is for reference)

There is nothing pending from my end, unless maintainers like to proceed further.

EmmmaTech:
fair

EmmmaTech:
why did you resolve my comments though?

run-test.sh
rsb-23: to keep things modular. 
Also, `run.sh` is coupled to timing function. So, it would be better to have `run` and `run-test` both call `timeit.sh`

run-test.sh
rsb-23: Yes it can be simplified later (in separate PR), once the workflow starts running smoothly.
As of now, merging the 2 seems pre-optimization and may impact existing design.

bddicken:
Hello there! This is cool, but at this point it is (a) a bit outdated since a ton of other languages have been added in and (b) I have since merged a PR that adds some git workflow / automation. I am going to close this one out, but if you have ideas for how to improve what's in here now, please feel free to open up another one. Thanks!

PR #72, opened 2024-11-26
Workflow for running tests on Linux (ubuntu)

Alvov1:
Adapted the compile.sh file to run inside the Github workflow with calling run.sh
Runs on ubuntu-latest (linux)


bddicken:
I think the github actions environment is too "unpredictable" for running the benchmark. These should be executed in a more controlled environment ideally. Going to close this one out.

PR #74, opened 2024-11-26
Enable ruby yjit

lazaronixon:


EmmmaTech:
I agree. on my machine (macbook air, m2) the average for ruby on the fibbonaci tests went from 16.76 seconds all the way down to 2.02! its speed went from comparing to php (being slightly slower) to comparing to pypy (being slightly faster).

numbcoder:
> I agree. on my machine (macbook air, m2) the average for ruby on the fibbonaci tests went from 16.76 seconds all the way down to 2.02! its speed went from comparing to php (being slightly slower) to comparing to pypy (being slightly faster).

use times instead of each will be faster. see this pr: https://github.com/bddicken/languages/pull/23

PR #75, opened 2024-11-26
Use functions and list comprehensions in Python

alexandermorgan:
This pr uses functions in both scripts like pr #31 and list comprehensions in the loops script. Both are improved, the latter going about 4X faster. There are other more performant ways to optimize in python (e.g. #53), but this approach is the biggest benefit available from pretty normal syntax/methods.

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -7,8 +7,11 @@ def fibonacci(n):
     return 1
   return fibonacci(n-1) + fibonacci(n-2)
 
-u = int(sys.argv[1])
-r = 0
-for i in range(u):
-  r += fibonacci(i)
-print(r)
+def main():
+  u = int(sys.argv[1])
+  r = 0
+  for i in range(u):
+    r += fibonacci(i)
+  print(r)
+
+main()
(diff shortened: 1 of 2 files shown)


EmmmaTech:
why was the code abstracted to call a main function in both python files? it's not necessary since python can just interpret each line as-is. plus, you didn't even add the `if __name__ == "__main__"` typeguard commonly used with main functions.

alexandermorgan:
@EmmmaTech the code was put in functions to make all the variables local instead of global because Python works a lot faster with local variables.

EmmmaTech:
is it really that advantageous when there's only max 1 defined function between both files?

alexandermorgan:
Yeah, wrapping the code in functions is all pr 31 does and they say it sped it up 2X. I think it's less than that but still really good such a simple change.

bddicken:
Thanks for the contribution.

As for putting it inside a function: That was added in via a different PR.

As for list comprehensions: The idea here is to implement it with the same algorithm as all the other ones. If I allowed for something like this, then to be fair I'd have to open up all the other versions to special optimizations and algorithmic enhancements.

PR #76, opened 2024-11-26
Php use typehints for fair comparation

ilnytskyi:
Php function signature is missing types.
With types should be more fair comparation to C as interpreter shouldn't be guessing dynamic type and hit jit as well

### DIFF
--- a/fibonacci/php/code.php
+++ b/fibonacci/php/code.php
@@ -1,6 +1,6 @@
 <?php
 
-function fibonacci($n)
+function fibonacci(int $n): int
 {
     if ($n == 0) return 0;
     if ($n == 1) return 1;


goetas:
php type checks are done at runtime, this adds two additional opcodes to check in and out param type

ilnytskyi:
@goetas while type checks indeed add more opcodes, they also should help jit determine types and genereate better instructions.

here my results, but the difference seems to be insignificant on my php version
```
PHP 8.3.11 (cli) (built: Aug 30 2024 09:27:49) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.3.11, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.11, Copyright (c), by Zend Technologies
```

Selection_20241128_017
Selection_20241128_018

bddicken:
Nice, thanks!

PR #77, opened 2024-11-26
Improve fibonacci function for Go

gandarez:
This PR improves how fibonacci function is written and reduces by ~97% (_0.58s_) its completion time. I testes in a Macbook Pro M2 Max with 32gb ram.


i3abghany:
The primary purpose of the benchmark is to test recursion so you shouldn't convert it into an iterative implementation.

PR #78, opened 2024-11-26
Automatically run benchmark on GitHub with Hyperfine

george-gca:
Added support to run the benchmarks automatically on GitHub with [GitHub actions](https://github.com/features/actions) and using [Hyperfine](https://github.com/sharkdp/hyperfine) to run the benchmarks and generate reports. The reports are automatically generated as a markdown table in each benchmark base directory, as can be seen in [fibonacci/fibonacci.md](https://github.com/george-gca/languages/blob/hyperfine/fibonacci/fibonacci.md) and [loops/loops.md](https://github.com/george-gca/languages/blob/hyperfine/loops/loops.md). If you want to check how everything runs, check my [PR for my own main repository](https://github.com/george-gca/languages/pull/1) and how the action runs.

PS: I recommend squashing this PR before merging, so this ridiculous commit history with lots of tests doesn't get transferred to the main branch.


george-gca:
Also, I am having problems with the rust action. Don't know how to fix it, but maybe this can be fixed by a regular rust developer after merging it with main and opening as a bug.

bddicken:
Thanks for doing this. I'm labelling as on hold for now, as Ive mostly been focussed on getting new languages in an optimized.

george-gca:
I have a suggestion. What if we comment out the automatic run instead and set to run only manually? This way this can go to the main branch and also we can run occasionally to check for newer benchmark results.

bddicken:
Thanks for for submitting this, but this is not needed at the moment.

PR #39, opened 2024-11-26
Added C# Loop and Fibonacci benchmarks

Azrielx86:

PR #79, opened 2024-11-26
I just added C# and F# code for loops

jonas1ara:
This pull request introduces several changes to add support for C# and F# in the project. The changes include updating scripts, adding new code files, and modifying project configuration files.

### Script updates:
* [`clean.sh`](diffhunk://#diff-ae565074a59b1eb6a8d16d9e8d3a40f89726ebf02debba40343b5d3f098ba9dcR7-R8): Added commands to remove compiled C# and F# code.
* [`compile.sh`](diffhunk://#diff-24d024fa911cc01760d679370d8ea8711230f073fde3e721a59e292b825252c4R8-R9): Added commands to build C# and F# projects using `dotnet`.
* [`run.sh`](diffhunk://#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35dR28-R29): Added commands to run compiled C# and F# executables.

### Code additions:
* [`loops/csharp/code.cs`](diffhunk://#diff-e806e8e1a58dbcde7e7f49a86ba22049a1bc0993e33f6f6f7248b90f62f3dea1R1-R21): Added a C# program that performs nested loops and prints a result.
* [`loops/fsharp/code.fs`](diffhunk://#diff-d26a745d1446e0ba31d17a42c4db495eee78063eff5f5092fb02c31b34e0d2d0R1-R14): Added an F# program that performs nested loops and prints a result.

### Project configuration:
* [`loops/csharp/code.csproj`](diffhunk://#diff-782d9bb65e1e42394ade8c00ffedae16b77c03c543bce56fc2bb37cb04b2ec6dR1-R10): Added a project file for the C# program, specifying the target framework and output type.
* [`loops/fsharp/code.fsproj`](diffhunk://#diff-965a91b1616d4e45ec6f8bc7824e9d31ab8e9fbb3e67a6470270623c196e7db1R1-R12): Added a project file for the F# program, specifying the target framework and output type.

### Miscellaneous:
* [`loops/.gitignore`](diffhunk://#diff-a258697a9edac7e4c4826dad714eb67ba5824962853ff9888336cba5be4152beR5-R7): Updated to include build directories for C# and F# projects.


jonas1ara:
I wish you could try it on your M3 machine to add the results to the project README.md

Emptycz:
Could you add a second version that runs on .NET 9? I am actually curious what the difference would be (I also do not own M3 neither "newer" CPU).

jonas1ara:
Yes, I also have that curiosity.

neon-sunset:
Thank you for submitting this PR. It looks like there are already multiple submissions. Also, this implementation does not use top-level statements for C# or the latest TFM for either language.

loops/fsharp/code.fs
```diff
@@ -0,0 +1,14 @@
+﻿open System
+
+[<EntryPoint>]
+let main args =
+    let u = int args.[0] // Get an input number from the command line
+    let random = Random()
+    let r = random.Next(0, 10000) // Get a random integer 0 <= r < 10k
```
neon-sunset, line 7: These two lines can be replaced with
```suggestion
    let r = Random.Shared.Next(0, 10000)
```

loops/fsharp/code.fsproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
neon-sunset, line 5: Here it is best to specify `net9.0`. It is also a good idea to specify `<PublishAot>true</PublishAot>`. When using AOT compilation, it is necessary to use `Console.WriteLine` instead of `printfn` instead.

run.sh
```diff
@@ -25,3 +25,5 @@ run "PHP" "php ./php/code.php 40"
 run "R" "Rscript ./r/code.R 40"
 run "Python" "python3 ./py/code.py 40" 
 run "Dart" "./dart/code 40"
+run "C#" "dotnet ./csharp/code/code.exe 40"
+run "F#" "./fsharp/code/code.exe 40"
```
neon-sunset, line 29: .NET does not output `.exe` files on Unix systems.

jonas1ara:
No, I kept it simple to be fair with the other languages, I even refrained from including an AoT compilation

neon-sunset:
It is fair to use the tool correctly. Which in this case means NAOT compilation - it is a fully-fledged target advertised for small and simple console applications like this one and not a niche. Same applies to writing C# in top-level statement style and using `var` (which the entirety of ecosystem defaults to for more than a decade).

loops/fsharp/code.fs
```diff
@@ -0,0 +1,14 @@
...
+    let a = Array.zeroCreate<int> 10000 // Array of 10k elements initialized to 0
```
neon-sunset, line 8: `int array` here is inferred automatically

jonas1ara:
Done @neon-sunset

neon-sunset:
@jonas1ara  Thanks a lot!

compile.sh
```diff
@@ -5,3 +5,5 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
 dart compile exe dart/code.dart -o dart/code --target-os=macos
+dotnet build -o ./csharp/code ./csharp/code.csproj
```
yunusefendi52, line 8: Shouldn't this `publish` instead of `build`?

compile.sh
```diff
@@ -5,3 +5,5 @@ RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Car
...
+dotnet build -o ./fsharp/code ./fsharp/code.fsproj
```
yunusefendi52, line 9: Shouldn't this `publish` instead of `build`?

run.sh
```diff
@@ -25,3 +25,5 @@ run "PHP" "php ./php/code.php 40"
 run "R" "Rscript ./r/code.R 40"
 run "Python" "python3 ./py/code.py 40" 
 run "Dart" "./dart/code 40"
+run "C#" "dotnet ./csharp/code/code 40"
+run "F#" "./fsharp/code/code 40"
```
yunusefendi52, line 29: Do we have to add "NativeAot" in this? like "C# (NativeAot)" or "F# (NativeAot)"
I'am sure there is the difference between them and other PR does not use native aot #39

neon-sunset: The other two PRs do use it. It's a good default for console applications. Both AOT and JIT use the same compiler back-end (the differences in optimizations won't show up in such microbenchmark because it does not do anything sensitive to optimizations exclusive to JIT or AOT), and it's more convenient to use AOT as it produces more consistent numbers in the cases like this and obviously cheaper startup cost.

compile.sh
neon-sunset: Good catch

isaacabraham: I'm not 100% sure about that - I recall an old discussion about this, but that might have been about `dotnet pack` perhaps?

isaacabraham: Ah, that's been fixed as of .NET 8 so this will indeed use Release.

loops/fsharp/code.fs
```diff
@@ -0,0 +1,13 @@
+﻿open System
+
+[<EntryPoint>]
+let main args =
+    let u = int args.[0] // Get an input number from the command line
```
isaacabraham, line 5: Shouldn't need the `.` here.

```suggestion
    let u = int args[0] // Get an input number from the command line
```

loops/csharp/code.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+class Program
+{
+    static void Main(string[] args)
+    {
+        var u = int.Parse(args[0]);               // Get an input number from the command line
+        Random random = new Random();
+        var r = random.Next(0, 10000);            // Get a random integer 0 <= r < 10k
+        var a = new int[10000];                // Array of 10k elements initialized to 0
```
SvenLauterbach, line 10: since the values are only positiv, this could also be uint, maybe even ushort?

bddicken:
There are some other C# PRs that came in before this, so I am going to close out this one in favor of those. If you think F# should be added, please open a separate PR with that for both loops and fibonacci. Thanks for the contribution.

PR #80, opened 2024-11-26
feat: add Julia loop code

exepulveda:
Julia script for loops problem

It's assumed Julia lang is installed


elbert5770:
```using BenchmarkTools

function main(args::Vector{String})
    u::Int32 = parse(Int32, args[1])                
    r = Int32(rand(0:9999))                 
    a::Vector{Int32} = zeros(Int32, 10000) 
    j::Vector{Int32} = collect(0:99999)              
    a .= sum(mod.(j, u)) .+ r
    println(a[r+1])                        # Print element (r+1 because Julia is 1-based)
end

@benchmark main($ARGS)

**Comment by exepulveda** [comment]
I assumed the idea is to do the same double for loop similar to all other languages

**Comment by ghyatzo** [comment]
Make sure to cast the integer to Int32, Rust and C both are using those, and the mod operation is considerably faster on 32 bit integers.

```julia
function main()
    u = parse(Int32, ARGS[1])
    r = rand(1:10000) % typeof(u)
    a = zeros(typeof(u), 10000)
    for i in 1:10000
        for j in Int32(1):Int32(100000)
           a[i] += j%u
        end
        a[i] += r
    end

    println(a[r])
end
```

also, you should compile it using `juliac` to avoid startup time.

bddicken:
Thanks for the contribution. We got Julia added in via a separate PR, so I'm gonna close this one out.

PR #83, opened 2024-11-26
Update README: Add explanation for script execution times

### LINKED ISSUE #64: Readme should include whether a higher or lower score is better.
While lower scores might seem intuitive, just by reading the readme whether higher or lower scores are more desirable is up to the interpretation of the reader. Clarification would make the project much simpler to understand.

James-Hou22:
- Added an explanation of the numbers output by `run.sh` to clarify what they represent.
- Highlighted that lower numbers indicate better performance.
- Clarified that only the third run is outputted by the script.

Closes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ To run one of the benchmarks:
 
 The numbers represent the real execution time (wall-clock time) it took for each language to execute the given task. **A lower number indicates better performance.**
 
-`bash ../run.sh` runs each program three times using the `runOnce` function and `awk` captures the real execution time; however, only the third run is recorded and outputted.
+`bash ../run.sh` runs each program three times using the `runOnce` function and `awk` captures the real execution time.
 
 ## Adding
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,12 @@ To run one of the benchmarks:
 
 4. For good measure, execute `$ bash ../clean.sh` when finished.
 
+### Interpretation
+
+The numbers represent the real execution time (wall-clock time) it took for each language to execute the given task. **A lower number indicates better performance.**
+
+`bash ../run.sh` runs each program three times using the `runOnce` function and `awk` captures the real execution time; however, only the third run is recorded and outputted.
```
EmmmaTech, line 34: https://github.com/bddicken/languages/blob/7d2d3ee6691335456a499975cfb2b1e9fe4937da/run.sh#L7-L12

this is incorrect, all three runs are recorded and outputted.

James-Hou22: Oops, I just noticed this as well. I got a little jumbled up as i was reading the readme because it was only showing one number. I'll fix this right now.

EmmmaTech (approved):
looks good :3

guibranco (approved):

bddicken:
Nice, thanks!

PR #84, opened 2024-11-27
Improve python loop performance by a couple percent

veganaize:
Use `+=` in python loop to slightly improve performance.

Maybe it'll beat the R implementation now?!


bddicken:
Thanks for the contribution.

PR #85, opened 2024-11-27
wrote batch scripts of clean, compile and run.

aldhinn:


bddicken:
Thanks for doing this. I'm labelling as on hold for now, as Ive mostly been focussed on getting new languages in an optimized.

bddicken:
I've decided not to go with `.bat` scripts for now.

PR #86, opened 2024-11-27
optimize go loops with goroutine

dolanor:
I know it seems to confront the language on some similar code.
But the original code was not using int32 like other codebase anyway.

But clearly, Go selling point in performance is the concurrency. Removing that from Go in an algorithm that can be concurrent makes it less interesting. Because yes, on a single CPU machine, Go will be slower than more performant languages like rust or C, because they don't have the overhead of the runtime.


neon-sunset:
Other languages have even nicer concurrency/parallelism primitives (you can trivially partition the workload with `Parallel.For` in C#), but it would fail to convey the main message behind these microbenchmarks.

EmmmaTech:
even if go's selling point involves concurrency, it's unfair to the other languages being compared because none of them also use concurrency. the code in each language is supposed to be as equivalent as possible to the other languages, and this doesn't seem equivalent at all.

MurtadhaS:
> Other languages have even nicer concurrency/parallelism primitives (you can trivially partition the workload with `Parallel.For` in C#), but it would fail to convey the main message behind these microbenchmarks.

that is subjective, for me Go's parallelism is the easiest and the cleanest.

neon-sunset:
> that is subjective, for me Go's parallelism is the easiest and the cleanest.

How would you implement the following in Go? This is a complete program and all requests are sent in parallel.
```cs
using var http = new HttpClient {
    BaseAddress = new("https://wttr.in")
};

var cities = new[] {
    "New York",
    "London",
    "Tokyo",
    "Kyiv",
    "Paris",
    "Berlin"
};

var reqs = cities
    .Select(c => $"{c}?format=3")
    .Select(http.GetStringAsync);

var wths = await Task.WhenAll(reqs);

Console.WriteLine(string.Concat(wths));
```

dolanor:
The correct way would be:

```go
package main

import (
	"fmt"
	"io"
	"log"
	"net/http"
	"os"
	"strings"

	"golang.org/x/sync/errgroup"
)

func main() {
	cities := []string{
		"New York",
		"London",
		"Tokyo",
		"Kyiv",
		"Paris",
		"Berlin",
	}

	var eg errgroup.Group
	acc := make([]string, len(cities))
	for i, c := range cities {
		eg.Go(func() error {
			resp, err := http.Get("https://wttr.in/" + c + "?format=3")
			if err != nil {
				return err
			}
			defer resp.Body.Close()

			b, err := io.ReadAll(resp.Body)
			if err != nil {
				return err
			}
			acc[i] = string(b)

			return nil
		})
	}

	err := eg.Wait()
	if err != nil {
		log.Println(err)
		os.Exit(1)
	}

	fmt.Println(strings.Join(acc, ""))
}
```

But since you don't handle exceptions, we could also do away with error handling
```go
package main

import (
	"fmt"
	"io"
	"net/http"
	"strings"
	"sync"
)

func main() {
	cities := []string{
		"New York",
		"London",
		"Tokyo",
		"Kyiv",
		"Paris",
		"Berlin",
	}

	var wg sync.WaitGroup
	acc := make([]string, len(cities))
	wg.Add(len(cities))
	for i, c := range cities {
		go func() {
			resp, _ := http.Get("https://wttr.in/" + c + "?format=3")
			b, _ := io.ReadAll(resp.Body)
			acc[i] = string(b)
			wg.Done()
		}()
	}

	wg.Wait()
	fmt.Println(strings.Join(acc, ""))
}
```

neon-sunset:
Thanks, my goal was to demonstrate @MurtadhaS that the only way to consider Go approach to concurrency superior is to be unfamiliar with better options. Goroutines are not composable and do even worse at ad-hoc fine-grained concurrency (like interleaving two concurrently sent requests - in .NET you simply skip calling `await` (or skip using `let!` binding in F#) and await later on when you do need the result of the task). Small nit - while my code does not handle exceptions, it does not silently ignore errors - if one or more requests fail, awaiting the task returned by `WhenAll` will throw `AggregateException` with the inner exception collection which can be acted upon as appropriate.

MurtadhaS:
> > that is subjective, for me Go's parallelism is the easiest and the cleanest.
> 
> How would you implement the following in Go? This is a complete program and all requests are sent in parallel.
> 
> ```cs
> using var http = new HttpClient {
>     BaseAddress = new("https://wttr.in")
> };
> 
> var cities = new[] {
>     "New York",
>     "London",
>     "Tokyo",
>     "Kyiv",
>     "Paris",
>     "Berlin"
> };
> 
> var reqs = cities
>     .Select(c => $"{c}?format=3")
>     .Select(http.GetStringAsync);
> 
> var wths = await Task.WhenAll(reqs);
> 
> Console.WriteLine(string.Concat(wths));
> ```

I mean the two code are comparable .
the biggest difference is the sheningans http.GetStringAsync is doing in C#, that makes the code looks smaller.
In addition to that, I like the verbosity of Go code.

bddicken:
Thanks for the contribution.

As for int32, that was added in via different PR, so we are good to go there.

As for concurrency: If I allowed for concurrency here, then I'd have to open it up to all the other languages as well. For the sake of this particular benchmark, no concurrency allowed (unless it's literally a special concurrent-only language or something unusual).

PR #88, opened 2024-11-27
Match Go integer size to C/Rust/Java integer size

zamadatix:
Go's int type will default to 64 bit on 64 bit machines, making it seem slower compared to other languages where 32 bit int is the default on modern machines.

Explicit integer sizes should probably be used on other languages as well for consistency and reproducibility.

There are other generated code output changes due to language differences but this is the lion's share and 2nd place seems to be the GC which is much more representative of expectations.

### DIFF
--- a/loops/go/code.go
+++ b/loops/go/code.go
@@ -9,11 +9,11 @@ import (
 func main() {
   input, e := strconv.Atoi(os.Args[1]) // Get an input number from the command line
   if e != nil { panic(e) }
-  u := int(input)
-  r := int(rand.Intn(10000))           // Get a random number 0 <= r < 10k
-  var a[10000]int                      // Array of 10k elements initialized to 0
-  for i := 0; i < 10000; i++ {         // 10k outer loop iterations
-    for j := 0; j < 100000; j++ {      // 100k inner loop iterations, per outer loop iteration
+  u := int32(input)
+  r := int32(rand.Intn(10000))           // Get a random number 0 <= r < 10k
+  var a[10000]int32                      // Array of 10k elements initialized to 0
+  for i := int32(0); i < 10000; i++ {         // 10k outer loop iterations
+    for j := int32(0); j < 100000; j++ {      // 100k inner loop iterations, per outer loop iteration
       a[i] = a[i] + j%u                // Simple sum
     }
     a[i] += r                          // Add a random value to each element in array


Trust-genie:
Pardon me but isn't it safer to have the compiler convert the type to the machines bit width at compile time? Running 32 bits on a 64bit machine and vice versa introduces all sorts of extra steps like bit padding and multiple computations since the bit width do not align.

zamadatix:
It's generally less "safe" as then your code has different numeric limits and limitations depending what machine you run it on. E.g. some builds will accept a certain value from the CLI and others won't.safest is to specify the size of the number you want to work and let the compiler figure out how to best accomplish that on the compilation target. The current default is "whatever the programming language thinks is most commonly useful" rather than "what the compiler thinks is most performant".

As far as performance, which is what I think you were meaning, smaller isn't a problem. The 64 bit version takes about 1.5x as much time as the 32 bit version in this patch on a 64 bit machine for example.

zamadatix:
Per the prior message, the current master takes 1.5x as long to execute as this patch when run on my 64 bit dev machine. More precisely:

time ./original 15427     -> yields ~1.8s
time ./edited15427      -> yields ~1.2s

I don't have an M3 to do a 100% direct comparison with the main results on the site though.

SerJaimeLannister:
What is your thoughts on the concurrent implementation proposed by https://github.com/bddicken/languages/pull/86 and how its benchmark compares since it also uses 32 bit

zamadatix:
Hmm, ultimately that's up to what you want this collection to represent. "1:1 comparisons (as much as the given language reasonably allows) of doing the same things" or "most optimized way to complete a given task in a given language someone is motivated to go make".

No right/wrong answer in my mind. Personally though, I'd lean towards the former. That a language e.g. makes concurrency simple and safe is, to me, more a benefit for the programmer and not necessarily the performance of the language between multiple sample cases. Plus the latter tends to make these competitions of who can be bothered to implement the cleverest algorithm and tweaks in their given language instead of useful comparisons.

ldemailly:
on my machine (m3 Mac) the 32 bit version is actually slower ... somehow

java 0.6s
go "int" 1.1s
go "int32" 1.9s
(for ref: c 0.56s)

zamadatix:
It's very possible the Go compiler is bad at targeting 32 bit on Apple Silicon vs Clang. E.g. whatever machine was used in https://github.com/bddicken/languages/pull/44 sees it as 3x faster, I see it as 1.5x faster on an 13th gen Intel machine, and Idemailly sees it as 1.7x slower.

This does lead to yet more open questions for the goals of the benchmark: Is code that is slower on 90% of machines but faster on the one test machine used because it's faster on the test machine or not used because it's slower in general? Does each language get to choose the fastest version of different bits or is the benchmark looking for how the language performs on a specific test, not how it forms on the fastest related test? What if the C version were hypothetically faster with a compiler that still interpreted int as uint16_t, would that compiler be faster or disqualified for modifying the test? I'm not sure many of these have a "right" answer either. The only thing I am sure of is the Go version does not match the other versions.

Unfortunately for the M3 case in particular since I don't have an M3 Mac to compare assembly between the Go and C versions with I'm unable to look into any of the assumptions of why 32 bit is faster for C but slower for Go on an M3 (and unable to test individual variations to see what the "ultimate combination" is on an M3, if that were even of interest in the benchmark).

zamadatix:
I realized I have access to an M2 at work, interestingly the edited version runs at the same speed between the original and edit for me on macOS 15.1.1 using go 1.23.3 darwing/arm64:

```
root$ time ./edited 5427
267706839

real	0m1.859s
user	0m1.841s
sys	0m0.018s

root$ time ./original 5427
267700919

real	0m1.871s
user	0m1.853s
sys	0m0.017s
root$ go version
go version go1.23.3 darwin/arm64

root$ cat original.go
package main
import (
    "fmt"
    "math/rand"
    "strconv"
    "os"
)

func main() {
  input, e := strconv.Atoi(os.Args[1]) // Get an input number from the command line
  if e != nil { panic(e) }
  u := int(input)
  r := int(rand.Intn(10000))           // Get a random number 0 <= r < 10k
  var a[10000]int                      // Array of 10k elements initialized to 0
  for i := 0; i < 10000; i++ {         // 10k outer loop iterations
    for j := 0; j < 100000; j++ {      // 100k inner loop iterations, per outer loop iteration
      a[i] = a[i] + j%u                // Simple sum
    }
    a[i] += r                          // Add a random value to each element in array
  }
  fmt.Println(a[r])                    // Print out a single element from the array
}

root$ cat edited.go
package main
import (
    "fmt"
    "math/rand"
    "strconv"
    "os"
)

func main() {
  input, e := strconv.Atoi(os.Args[1]) // Get an input number from the command line
  if e != nil { panic(e) }
  u := int32(input)
  r := int32(rand.Intn(10000))           // Get a random number 0 <= r < 10k
  var a[10000]int32                      // Array of 10k elements initialized to 0
  for i := int32(0); i < 10000; i++ {         // 10k outer loop iterations
    for j := int32(0); j < 100000; j++ {      // 100k inner loop iterations, per outer loop iteration
      a[i] = a[i] + j%u                // Simple sum
    }
    a[i] += r                          // Add a random value to each element in array
  }
  fmt.Println(a[r])                    // Print out a single element from the array
}
```

So if there really is a slowdown on Apple Silicon it'd appear to by hyper-specific to the M3 for some odd reason. It's also possible my Go and/or macOS versions are different than Idemailly's and that's the source. Hard to say without having specifically an M3 to test.

d0ngw:
The benchmark result for go1.22.3 darwin/arm64 is much faster if the loop order is swapped.

```
real	0m0.675s
user	0m0.601s
sys	0m0.004s
```
```
package main

import (
	"fmt"
	"math/rand"
	"os"
	"strconv"
)

func main() {
	input, e := strconv.Atoi(os.Args[1]) // Get an input number from the command line
	if e != nil {
		panic(e)
	}
	u := int(input)
	r := rand.Intn(10000) // Get a random number 0 <= r < 10k
	var a [10000]int      // Array of 10k elements initialized to 0

	for j := 0; j < 100000; j++ { // 100k inner loop iterations, per outer loop iteration
		for i := 0; i < 10000; i++ { // 10k outer loop iterations
			a[i] = a[i] + j%u // Simple sum
			if j == 0 {
				a[i] += r // Add a random value to each element in array
			}
		}
	}
	fmt.Println(a[r]) // Print out a single element from the array
}

```

gmr458:
@zamadatix My laptop specs:

Intel Core i3 6006U
8GB RAM 2133MHz

I used 40 as input, 15 seconds using `int`, 5 seconds using `int32`.

bddicken:
Thanks for the contribution!

We got this added in in another PR, so going to close this one out.

ldemailly:
looks like this was in #44

PR #89, opened 2024-11-27
Make python fibonacci loop one-based, like the other languages

veganaize:
The other language implementations have a fibonacci `for` loop initialized to `1`...

The python `range()` function will default to `0` when no start value is specified; it will stop at size-1 no matter what.  (Effectively `i < u` in the other languages).

This change makes the python implementation's fibonacci `for` loop iterate over `1` through `u-1`, just like the other languages do.


ximega:
The difference in performance is just one call of function,
 where it straightaway returns 0 with only ONE if-statement

veganaize:
> The difference in performance is just one call of function

Yes, it is...  But shouldn't it be consistent with the other language implementations?

> [...] it straightaway returns 0 with only ONE if-statement

It should behave just like the other languages.  (I'm not sure I follow?)

bddicken:
Good catch, thanks for contributing.

PR #91, opened 2024-11-27
Use pointer arithmetic in C

bartash:
Use pointer arithmetic to gain a little speed, maybe 10%


bddicken:
Thanks for submitting this. Did you run before / after tests? I'd be somewhat surprised if this makes much of a difference with -O3 enabled.

bddicken:
Closing this out as I have not hear back in a week. Feel free to re-open if you wanna add some numbers.

PR #92, opened 2024-11-27
Optimize Python Code with Numba JIT

is9117:
Optimized the test code using **Numba's Just-In-Time (JIT) compilation**, reducing execution time from **74 seconds** to **2.4 seconds**.

## Changes

- **Added Numba Import**: `from numba import njit`
- **Applied `@njit` Decorator** to the `run` function to speed up nested loops.

## Code

```python
import sys
import random
from numba import njit

@njit
def run(u, r):
    a = [0] * 10000
    for i in range(10000):
        for j in range(100000):
            a[i] += j % u
        a[i] += r
    print(a[r])

u = int(sys.argv[1])
r = random.randint(0, 9999)
run(u, r)
```

## Notes

- **Numba JIT** compiles Python code to machine code at runtime, greatly improving performance with minimal changes.


EmmmaTech:
what's the point of this? pypy has jit compilation by default, and both pypy and cpython (the default python implementations) are tested in the codebase.

EmmmaTech:
nevermind, I realized that the code here is in a different folder. I somehow thought you were modifying the original python code 😭

bddicken:
Thanks for submitting this. Though, I'm not sure this is necessary since we already have a JIT version for python (PyPy). Is there a particular reason why this should be added in addition to PyPy?

bddicken:
We are getting to the point where the number of distinct languages is very high. I'm going to close this out and just keep python3.13 and pypy.

PR #93, opened 2024-11-27
Add Scala

jpablo:
It needs `scala-cli` to compile:

https://scala-cli.virtuslab.org/install

```
brew install Virtuslab/scala-cli/scala-cli
```


bddicken:
Nice, thanks for contributing.

PR #94, opened 2024-11-27
allocte array with provided size in go

Anupam-Ashish-Minz:
Changes the performance of go roughly by a factor of 10.


Anupam-Ashish-Minz:
You are right, I didn't check with go build, I just ran with go run. My bad I'll fix it.

Anupam-Ashish-Minz:
This was a false flag, I clearly shouldn't have done it while being half asleep. I totally missed the errors the program was throwing and went with the output. Thus I am closing this issue.

PR #97, opened 2024-11-27
Enable JIT and disable XDebug for PHP

CViniciusSDias:
I enabled JIT and disabled XDebug on PHP to make the benchmark a bit more fair. A debugger can severely slow down the execution and other languages have JIT enabled.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -11,8 +11,6 @@ function run {
   runOnce "$1" "$2"
 }
 
-cd loops
-
 run "Kotlin" "java -jar kotlin/code.jar 40"
 run "C" "./c/code 40" 
 run "Go" "./go/code 40" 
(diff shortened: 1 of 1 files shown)


Hattorius (approved):
Good job frien

WhiteCranee:
The run.sh is written so you cd into the test you want to perform either you cd into loops or fibonacci and then execute the run.sh from there so it performs the commands relativ to your current directory.

CViniciusSDias:
I missread the README.md, @WhiteCranee . I read `./run.sh` instead of `../run.sh`. Thank you. I removed it.

bddicken:
Thanks for the contribution.

Jit was added in a separate PR, however.

PR #99, opened 2024-11-27
Use enumerate for outer python `for` loop

veganaize:
Using `enumerate()` instead of `range()` is a very common practice in python when both the index and the value will be utilized.

This type of idiom affords readability improvements, as well as a slight speed optimization, over the `a[i]` style.

_(This PR offers even better performance than the solitary `+=` optimization in my #84 PR)_


veganaize:
The other, statically compiled implementations are highly likely to optimize their array accesses behind the scenes, especially when compiled with optimization flags set.  If we are going to be fair, those language implementations should be compiled with nothing other than the default optimization level (ie. no additional flags set at compile / link time).

While I completely understand `enumerate` is considered idiomatic, I believe it holds here and other languages should be allowed to do something very similar if it does in fact offer an improvement in performance.  (Which I don't think it will for the languages compiled with optimizations.)

bddicken:
I'm going to stick with the existing way the program is implemented for the time being.

PR #103, opened 2024-11-27
Add Crystal

Tamnac:
Partially based off of ruby implementation.

For fibonacci, it's 2x as fast to use wrapping math. I'm not sure whether we are supposed to, but since the stated purpose is to test function calls, not math, I'm going to assume yes.


bddicken:
Crystal seems cool! Thanks for the contribution.

PR #104, opened 2024-11-27
Add Odin

AyeAreEm:
Added odin to `fibonacci/` and `loops/`


guaracy:
i think you can try

- compile with `odin build odin/code.odin -o:speed -file` to produce faster executables

- in fibonacci, you can change the two lines

```odin
if n == 0 do return 0
if n == 1 do return 1
```
by
```odin
if n < 2 do return n
```

SerJaimeLannister:
any benchmark ?

bddicken:
Nice, thanks for contributing.

PR #81, opened 2024-11-26
Added ADA

gab5987:
Created benchmarks for the ADA programming language.

PR #105, opened 2024-11-27
Added Ada tests

iharuspex:
Added Ada tests


Irvise:
Hi @iharuspex, as a fellow Ada programmer, thanks for adding Ada ^^

A couple of recommendations. With how simple the code examples are (a single file), I think that using Alire is overkill. A simple `gnatmake` would suffice. Which takes me to the second point.

Alire, by default builds things in development mode unless told otherwise, which you did by adding the `--release` flag to `alr`. However, that still does not enable all possible performance flags for the compiler. For example, the Rust implementation enables `-lto` and `-strip`, which Alire does not enable. And Alire's performance defaults to `-O2` instead of `-O3`. It also leaves some runtime checks.

What do you think about dropping Alire and GPRbuild and just defaulting to GNATmake plus flags? This method also reduces the amount of dependencies that people may need to install in their computers to test it themselves :)

Best regards,
Fer

Irvise:
Also, I just saw #81 :S

Irvise:
In the loop example you are doing a `rem` operator. I believe it has to be the `mod` operator. I would also put that expression in parenthesis for good measure.

Your code is basically the same as #81, but with nicer compiler flags :)

There are a couple of things that I would change, but you really do not need to do them. Consider them as an opinion, not a suggestion:
- For such a simple example, I would not use the `@` operator, as newer people may be confused by it and it does not save any space here
- For Fibonacci I would use `Natural` or `Positive` as an input type, as it does not make sense to receive negative values.
- Recursive declaration of functions (Fibonacci) is fine most of the time. But with large inputs (huge recursion trees) and when performance is needed, I would always turn them into a simple `for ... loop` with an accumulator. This is what tail recursion is in Scheme/Lisp and other languages :)

Best regards,
Fer

iharuspex:
`rem` and `mod` work the same with positive dividend, `rem` works basically the same as `%` in C. I put this expression in parentheses for better readability.

I agree that using `@` here may not be very justified, but it improves readability a little, in my opinion. Anyway, I decided to  replace it with an expression like `X := X + Y` for better clarity.

Also, I agree that using `Natural` type in the Fibonacci test would look much better :)

As far as I understand, recursive calculation of Fibonacci number is done here only for performance testing. Of course, it is not the most efficient algorithm :)

Thank you for your code review and advice!

Best regards,
Artem

bddicken:
Awesome, thanks for the contribution.

PR #107, opened 2024-11-27
Add GraalVM native image as option to run Java.

thomaswue:
This adds the capability to run the Java code as a GraalVM native image. For installing GraalVM, one can use SDKMan:
`sdk install java 23.0.1-graal`

Alternatively, the binaries are available here: [https://www.graalvm.org/downloads/](https://www.graalvm.org/downloads/)

Measurements on my machine (Intel 13th Gen Core i9-13900K) show that it brings the time for Java for the loop example down to the levels of C/C++ and Rust: 
```
C = 1.52
C = 1.50
C = 1.51

Java = 1.59
Java = 1.60
Java = 1.60

Java GraalVM = 1.51
Java GraalVM = 1.50
Java GraalVM = 1.50
```


helpermethod:
I wonder if it would be even better to generate the random number at build time 😄

thomaswue:
@helpermethod With proper loop transformations, the primary work of the benchmark (summing up the modulo values) could be done outside the outer loop. This would then reduce the time to almost zero. We have all the relevant optimizations available in the Graal compiler, but the current heuristics do not play well enough together to trigger the transformation.

bddicken:
Awesome, thanks.

PR #109, opened 2024-11-27
added cpp

Aparup123:
Added the code for `C++`. Also added the code for `C++ `in `compile.sh`, `run.sh` and `clean.sh` files

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,4 +1,5 @@
 clang -O3 c/code.c -o c/code
+clang++ cpp/code.cpp -o cpp/code
 go build -o go/code go/code.go
 javac jvm/code.java
 RUSTFLAGS="-Zlocation-detail=none" cargo +nightly build --manifest-path rust/Cargo.toml --release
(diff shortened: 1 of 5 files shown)


Hxnter999:
How is this any different or any better than the C version of the code? at least take the time to rewrite using modern C++ features.

Aparup123:
> How is this any different or any better than the C version of the code? at least take the time to rewrite using modern C++ features.

Sorry, i followed the same style cause then only we can compare the languages for the same code.

compile.sh
```diff
@@ -1,4 +1,5 @@
 clang -O3 c/code.c -o c/code
+clang++ cpp/code.cpp -o cpp/code
```
sergev, line 2: Flag ```-O3``` would help a little bit.

bddicken:
Thanks for the contribution. Technically, C++ is a superset of C, and what we have here is basically just C code. Since it also uses the same complier (clang) this is basically just going to be the same thing. Therefore, I'm not going to include it.

Team C++ can just use the C scores.

PR #110, opened 2024-11-27
Add AssemblyScript - running on Bun

WebReflection:
This MR provides an [AssemblyScript](https://www.assemblyscript.org/) implementation, currently running on the fastest JS runtime available around this project.


SerJaimeLannister:
Interesting but I am not sure what this is for ?

You have written something like typescript which compiles to web assembly which you then run in bun.

I honestly have seen so many webassembly runtimes on websites like hackernews which is why it is so interesting as to why you are using bun though

https://wasmer.io/ 
This is what came at the top of my mind but there are many others
https://00f.net/2023/01/04/webassembly-benchmark-2023/
https://github.com/bytecodealliance/wasm-micro-runtime

And have you benchmarked it , is it faster than the current bun implementation ? 

I am genuinely curious about this

WebReflection:
The point is to measure WASM instead of interpreter and you need a reference benchmark to see the variation from it … Bun is already used in this benchmark, seeing how WASM in bun can improve or degrade performance is, imho, interesting. Using any other WASM runner would do too but the point of WASM is its ability to tune everywhere but there is also a myth about its performance which I expect to be lower in these two test cases, not better than JIT. Although I don’t have /urn/bin/time on Linux so I wouldn’t know

SerJaimeLannister:
You could try something like hyperfine for benchmarking it locally and present your results , if maybe the author / general people like it , they can add it to the main repo

https://github.com/sharkdp/hyperfine

atleast that is what I have done in the lua implementation that I have provided because I am much more familiar with hyperfine

SerJaimeLannister:
hyperfine --warmup 3 \
          --min-runs 5 \
          --show-output \
          "command here"

Here is an example of hyperfine which I had used to benchmark the lua example created by me

WebReflection:
```
hyperfine --warmup 3 --min-runs 5 --show-output 'bun ./as/code.js 40'
Benchmark 1: bun ./as/code.js 40
1952033
1952033
1952033
1952033
1952033
1952033
1952033
1952033
  Time (mean ± σ):      3.126 s ±  0.026 s    [User: 3.120 s, System: 0.008 s]
  Range (min … max):    3.113 s …  3.173 s    5 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

hyperfine --warmup 3 --min-runs 5 --show-output 'bun ./js/code.js 40'
Benchmark 1: bun ./js/code.js 40
1958295
1956754
1951443
1959577
1950172
1955526
1957070
1959795
  Time (mean ± σ):      1.338 s ±  0.001 s    [User: 1.332 s, System: 0.009 s]
  Range (min … max):    1.337 s …  1.339 s    5 runs
```

WebReflection:
Passing the random number from the runner it removes the warning but the result is similar:

```
hyperfine --warmup 3 --min-runs 5 --show-output 'bun ./as/code.js 40'
Benchmark 1: bun ./as/code.js 40
1958388
1953070
1950586
1958709
1952353
1958227
1950124
1959085
  Time (mean ± σ):      3.150 s ±  0.005 s    [User: 3.139 s, System: 0.012 s]
  Range (min … max):    3.144 s …  3.155 s    5 runs
```

WebReflection:
**fibonacci**

```
hyperfine --warmup 3 --min-runs 5 --show-output 'bun ./as/code.js 40'
Benchmark 1: bun ./as/code.js 40
165580140
165580140
165580140
165580140
165580140
165580140
165580140
165580140
  Time (mean ± σ):      1.132 s ±  0.003 s    [User: 1.122 s, System: 0.012 s]
  Range (min … max):    1.130 s …  1.136 s    5 runs
 
hyperfine --warmup 3 --min-runs 5 --show-output 'bun ./js/code.js 40'
Benchmark 1: bun ./js/code.js 40
165580140
165580140
165580140
165580140
165580140
165580140
165580140
165580140
  Time (mean ± σ):     913.0 ms ±   1.8 ms    [User: 902.4 ms, System: 11.6 ms]
  Range (min … max):   910.0 ms … 914.5 ms    5 runs
```

I think it's interesting seeing WASM outperforming PHP, Python (not PyPy), and Ruby, but I let you decide what to do with this PR.

bddicken:
Thanks for putting in the effort to make this.

Can you provide a good argument for why this is needed? Based on your posts, it seems like the performance of "regular" bug is much better then the AssemblyScript versions. What exactly does this add knowledge-wise?

WebReflection:
would it make more sense if the generated WASM would run via PHP or Python to you? In there we should see how faster WASM can be

bddicken:
Thanks for putting this together, however I feel as if this is beyond the scope of the project at this time.

PR #115, opened 2024-11-27
Added compiled bun to the tests

pacosw1:
Bun can compile into a binary to run the code without dependencies.

Added a compiled bun test.

Tested on my local machine


bddicken:
Cool, thanks.

PR #118, opened 2024-11-27
fixes : #111

### LINKED ISSUE #111: Add a benchmark score result table or chart
Can we add a table or a chart to show all the benchmark scores for each language, in different sorts of environments, maybe it would be more intuitive.

AshrithSathu:
displaying benchmark visualzation in terminal

fixes #111 

benchmark


AshrithSathu:
@bddicken This is more of an improvement than an issue, but I found a issue open with the same thing, so I tagged it. Please consider it as an improvement.

bddicken:
This is really cool. I'm mostly focussing on gettin new languages and implementation improvements in for now, so I'll label this as `on hold` for the time being.

AshrithSathu:
Sure @bddicken will update all the languages in the end

bddicken:
The benchmark has evolved quite a bit since this was first created. While it is a cool idea, I'm going to close this out so we can clean up some of the issues and PRs.

PR #120, opened 2024-11-27
Add Nim

Beetroit:
Added code for nim in fibonacci and loops.
https://nim-lang.org/install_unix.html


m4ul3r:
https://github.com/bddicken/languages/pull/41

As far as the loop example, I think the earlier PR is more accurate to C. The use of newSeq will allocate in the heap, should use an array to keep it as comparable.

bddicken:
Super cool, thanks.

PR #122, opened 2024-11-27
F95 impementation

exepulveda:
Implementing loop challenge in >= Fortran 95


bddicken:
Fortran has been added at this point. Does fortran 95 add any significant improvement over whats in the repo already?

PR #125, opened 2024-11-28
Fix random seed in C

EienMiku:
`rand()` is used without `srand()` above, `rand()` behaves as if it was seeded with `srand(1)`, which generates the same value each time.


bddicken:
Nice, thanks.

PR #126, opened 2024-11-28
AdD

cyrusmsk:
Simple D implementation, without tweaks on types, TCO, simd and other possible stuff


bddicken:
Cool, thanks for contributing.

PR #128, opened 2024-11-28
Updated R's Fibonacci function

Shea-Fyffe:
The new code should be much faster. Changes made:

- Removed the loop for summing; it's never a good idea to loop to sum a vector of values (we can just use `sum()`)
- Removed recursion from the `fibonacci()` function and added a pre-allotted vector
- Could be improved further with memoization

### DIFF
--- a/fibonacci/r/code.R
+++ b/fibonacci/r/code.R
@@ -1,21 +1,21 @@
-# Recursive Fibonacci function
+# Updated Function with Pre-alloted vector
 fibonacci <- function(n) {
-  if (n == 0) return(0)
-  if (n == 1) return(1)
-  return(fibonacci(n - 1) + fibonacci(n - 2))
+    fib <- vector("integer", n)
+    if (n < 2) {
+        return(n)
+    }
+    fib[1:2] <- c(1, 2)
+    for (k in 3:n) {
+        fib[k] <- fib[k - 1] + fib[k - 2]
+    }
+    return(fib)
 }
-
 # Parse command-line argument for `u`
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for the contribution.

The idea is not to create the most highly-optimized solution to solve the problem, but to create an implementation that is the "same" across all languages. If I allowed for these types of optimizations, then to be fair, I'd have to open up all the other implementations to optimizations such as tail-recursive solutions or iterative solutions.

Shea-Fyffe:
> Thanks for the contribution.
> 
> The idea is not to create the most highly-optimized solution to solve the problem, but to create an implementation that is the "same" across all languages. If I allowed for these types of optimizations, then to be fair, I'd have to open up all the other implementations to optimizations such as tail-recursive solutions or iterative solutions.

IMO that should be made more explicit considering the major benchmark is about optimization, and "the same" code being used is far more suboptimal for some languages than others. Moreover, there are nuances in the examples that make them not truly similar (e.g., using `switch()` instead of `if/else`, using `+=` instead of `obj = obj + ...`). 

If your goal is to apply to examine how fast languages execute particular boiler-plate code that is one thing. But I feel like it's being marketed as how fast languages execute certain tasks, which doesn't seem true.

People will see this and think Python, R, etc are garbage languages—especially if they don't do further investigation of the code.

Maybe a disclaimer should be provided that PRs shouldn't allow for improvements, and that the tasks involved executing the same boiler-plate code which some languages will perform worse by design.

After saying all that, I do appreciate the work you've done 🙏🏻

PR #129, opened 2024-11-28
Optimize python solution

thmasq:
I got some interesting optimization results. The execution time went from around `82` seconds on my machine (Ryzen 7800X3D), down to `31` seconds. I think it's worth looking into.

I also prepared some tests to show the results are the same
```python
import random
import sys

def optimized_computation(u):
    a = [0] * 10000

    r = random.randint(0, 10000)

    for i in range(10000):
        total = 0
        for j in range(100000):
            total += j % u
        a[i] = total + r

    return a[r]

def original_computation(u):
    r = random.randint(0, 10000)
    a = [0] * 10000
    
    for i in range(10000):
        for j in range(100000):
            a[i] = a[i] + j % u
        a[i] = a[i] + r
    
    return a[r]

def compare_computations(u, num_runs=5):
    for run in range(1, num_runs + 1):
        random.seed(run)
        result1 = optimized_computation(u)
        
        random.seed(run)
        result2 = original_computation(u)
        
        print(f"Run {run}:")
        print(f"  Optimized Method Result: {result1}")
        print(f"  Original Method Result:  {result2}")
        print(f"  Results Match: {result1 == result2}\n")

if __name__ == "__main__":
    u = int(sys.argv[1])
    compare_computations(u)
    ```

### DIFF
--- a/loops/py/code.py
+++ b/loops/py/code.py
@@ -1,10 +1,20 @@
-import sys
 import random
-u = int(sys.argv[1])         # Get an input number from the command line
-r = random.randint(0, 10000) # Get a random number 0 <= r < 10k
-a  = [0] * 10000             # Array of 10k elements initialized to 0
-for i in range(10000):       # 10k outer loop iterations
-  for j in range(100000):    # 100k inner loop iterations, per outer loop iteration
-    a[i] = a[i] + j%u        # Simple sum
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for your contribution. This may run faster, however it deviates from the way the other implementations are done by adding another variable and reducing the amount array accesses. Due to this, it won't be merged.

PR #130, opened 2024-11-28
Use Hyperfine for benchmarking

thmasq:
Hyperfine has many useful tools for this kind of benchmark, among them are running warm ups, disabling the shell for lower overhead, and detecting statistical outliers on measurements.

------

Untitled


bddicken:
This is awesome.

PR #133, opened 2024-11-28
Adding D benchmarks

pjmlp:
Examples for D language, using the [LDC](https://github.com/ldc-developers/ldc) variant of the compiler.


cyrusmsk:
Almost identical :) https://github.com/bddicken/languages/pull/126
I’ve also added no boundscheck and I think foreach is more idiomatic way for loops

bddicken:
Thanks for contributing! However, we got D added in from a different PR, so going to close this one out.

PR #135, opened 2024-11-28
Add Lua and LuaJIT

felipeguilhermefs:
Quick and easy add of Lua and LuaJIT.

Lua (5.4.7) beat all the other interpreted languages in my machine.
LuaJIT (2.1.17...) is mostly quicker than  Go. 

Code is pretty simple, so it should be able to run on older versions as well.


bddicken:
Thanks! Got Lua added in via https://github.com/bddicken/languages/pull/46 though, so I'm going to close this out.

PR #136, opened 2024-11-29
Rework of C loop for performance improvements, and fixed RNG

NoxMaw:
RNG needs time.h to work properly, fixed that.
Besides this, rewored the whole C code to improve performance:
- Using Registered unsigned integers of 32 bits instead of singed integers
- Avoid printf, instead use 'write' with a buffer
- Unwinded loop to reverse the integer in order for 'write' to print it out properly

### DIFF
--- a/loops/c/code.c
+++ b/loops/c/code.c
@@ -1,16 +1,63 @@
-#include "stdio.h"
-#include "stdlib.h"
-#include "stdint.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <unistd.h>												                                   // include this for the write function
+#include <time.h>                                                           // time.h added to seed rng
 
-int main (int argc, char** argv) {
-  int u = atoi(argv[1]);               // Get an input number from the command line
-  int r = rand() % 10000;              // Get a random integer 0 <= r < 10k
-  int32_t a[10000] = {0};              // Array of 10k elements initialized to 0
-  for (int i = 0; i < 10000; i++) {    // 10k outer loop iterations
-    for (int j = 0; j < 100000; j++) { // 100k inner loop iterations, per outer loop iteration
-      a[i] = a[i] + j%u;               // Simple sum
-    }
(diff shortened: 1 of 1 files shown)


NoxMaw:
Indeed, this code has a very small performance gain, but really 'optimize speed over everything else' is all I had in mind writing this hahah
I don't see a way to really improve the performance of this without modifying the nested loop itself, or the whole program for that matter, and that's not allowed.


> I think you're new to C so I'm not accusing you of concealing something here, but, what you left out in your PR comment is that you hardcoded `u = 40` instead of reading it from `argv`... which is literally only thing from your PR that actually makes a measurable difference...

Ah, that's what I get for writing code at 3 in the morning, sorry about that, it was unintentional.
And yeah I am quite new at C, I have been coding in C for about two years or so ^^

> only saves around 0.005ms (5 microseconds) on my machine (the benchmark is 1.099s, so it's several orders of magnitude too insignificant to justify a 4x loc increase)

I admit it is indeed a petty difference, there is very little to 'optimize' here, considering it's just a nested loop and the 100,000 % operations and constant memory access are the real performance hogs here.

> your formatting algorithm is wrong, it truncates one digit too early

It's actually a mistake in the 'write' function, printed write(; ; x), should have been  write(; ; x + 1) to print the full number, good catch!

>flags a warning because you don't check the result of write which has warn_unused_result

Indeed, this code is very unsafe, it will crash as soon as you pass something like a negative number to it or anything that is unexpected. This thing will only work as long as it keeps getting max_9_digits positive integers so it will not crash within the bounds of this small program.
I could have used putchar() for this instead of write (it's safer), but that would imply more syscalls and I'm saving as many CPU cycles as humanly possible here, so I went for the unsafe write with a buffer to ship it all on a single syscall.

>rand() sticking to the default seed of 1 is likely intentional, to ensure consistent results across runs (could be wrong, this is 100% speculation on my part)

I had not considered this, makes sense though, it would be cool to leave a comment specifying in the code.


And your comments about register and signed integers are very true, thanks for the tip, and all the feedback!

PR #137, opened 2024-11-29
Add explicit JIT buffer size for php

goetas:
php 8.3 needs this param, 8.4 not anymore (added in https://wiki.php.net/rfc/jit_config_defaults)


bddicken:
Nice, thanks.

Issue #114, opened 2024-11-27
jvm folder


ArtyomKingmang:
Why are Kotlin and jvm different folders? Kotlin is also a language for the jvm platform. You either need to add Kotlin to the jvm package, and add the Java code to another package inside the jvm. Or rename the jvm package to java

PR #138, opened 2024-11-29
fix package name

ArtyomKingmang:
Fixed the problem described in my issue: https://github.com/bddicken/languages/issues/114


bddicken:
Thanks! Need to also update the run script.

ArtyomKingmang:
everything seems to be

PR #139, opened 2024-11-29
add a temp variable for the sum

jersobh:
Creating a temporary variable to accumulate the results of j % u will improve performance as it will reduce the number of accesses to the array a[i].


jrosell:
It's true for most languages. And ones are more affectes than others.

bddicken:
Thanks for the idea here. I do understand that adding a temp variable can help improve performance, but this is a fundamental change to the program. It would either have to be all or nothing for all the languages. For the time being, I'm going to go with "nothing."

PR #140, opened 2024-11-29
Add Nim programming language

GabrielBigardi:
The code is as close as possible to the other codes


bddicken:
Thanks for contributing! We got nim in from a different PR, so I'm gonna close this one out.

PR #141, opened 2024-11-29
improve swift optimization

nervenes:
previously, the loops implementation was very fast at the cost of making the fibonacci implementation slow. the ``-Ounchecked`` flag optimizes better for fibonacci but made the loops slower, so i've implement a middle ground that keeps the loops implementation about similar in speed (about 0.01-0.02s slower) while improving the fibonacci implementation by +25% in speed.


fibonacci/swift/code.swift
```diff
@@ -1,8 +1,7 @@
 @main
 struct Main {
     static func fibonacci(_ n: Int) -> Int {
-        if n == 0 { return 0 }
-        if n == 1 { return 1 }
+        if n <= 1 { return n }
```
bddicken, line 4: We just need to revert this change.

bddicken:
I'm going to leave swift as-is for now, but thanks for submitting this idea.

PR #142, opened 2024-11-29
Kotlin

programadorthi:
> Previous PR #15 was closed because there were a lot of conflicts using main reference directly

Bring kotlin JVM, JS and Native.

JVM is working as previosly.
Native needs [download and install the Kotlin Native Compiler.](https://kotlinlang.org/docs/native-command-line-compiler.html#download-and-install-the-compiler)
JS almost of the time come with JVM compiler but missing the lib folder that contains que `kotlin-stdlib-js.klib`. So a [manual install](https://kotlinlang.org/docs/command-line.html#manual-install) would be better.
After that open `compile.sh` and set `KOTLIN_HOME` to Kotlin directory.


bddicken:
Thanks for submitting this. Why are there different implementations for kotlin native vs with the jvm? I'd like to just stick with the same code for all versions if possible.

beradeep:
Hi @bddicken, Kotlin officially provides both `jvm` and `native` targets. The implementations aren't different at all. The same code can be compiled to create `.jar`s for the `jvm` as well as native binaries like the other languages. The `kotlinc` compiler compiles to `jvm` targets while `kotlinc-native` compiles to native targets. So also including `native` to the test will perhaps give a fair representation of Kotlin imo.

beradeep:
Also, I think `kotlin-js` is unnecessary (kind of irrelevant :) ) here.

bddicken:
I went ahead and merged #171, so I will close this out. Thanks though for submitting.

PR #143, opened 2024-11-29
Adding vectorized Python code using numpy

Karatsidis:
## Rationale
Python's popularity is owed to the abundance of libraries such as numpy written in lower-level languages e.g. C and as a result boosting the Python performance in cases where vectorization is possible.
It would be interesting to understand how a vectorized Python solution using Numpy would perform compared to other languages and native Python. 

## Description of change
Code is provided for the two problems (loops and fibonacci) using vectorized solutions.

## Notes:
- additional requirement `numpy` to be installed via `pip` or other package manager.


cyrusmsk:
https://github.com/bddicken/languages/pull/63

loops/py/code.py
```diff
@@ -1,13 +1,16 @@
...
+    a = [0] * 10000             # Array of 10k elements initialized to 0
     for i in range(10000):       # 10k outer loop iterations
-      for j in range(100000):    # 100k inner loop iterations, per outer loop iteration
-        a[i] = a[i] + j%u        # Simple sum
-      a[i] = a[i] + r            # Add a random value to each element in array
-    print(a[r])                  # Print out a single element from the arra
+        for j in range(100000):    # 100k inner loop iterations, per outer loop iteration
+            a[i] = a[i] + j % u        # Simple sum
+        a[i] = a[i] + r            # Add a random value to each element in array
+    print(a[r])                  # Print out a single element from the array
```
bddicken, line 13: Please revert the changes to the regular `py` versions so we can keep this focused on just the `numpy` variants.

bddicken:
Thanks for the contributions. A few updates are needed here.

loops/py/code.py
Karatsidis: updated

run.sh
```diff
@@ -29,6 +29,8 @@ run "PHP" "php ./php/code.php 40"
 run "R" "Rscript ./r/code.R 40"
 run "Python" "python3.13 ./py/code.py 40"
 run "Common Lisp" "common-lisp/code 40"
+run "Python" "python3 ./py/code.py 40"
```
Karatsidis, line 32:
```suggestion
```

fibonacci/numpy/code.py
```diff
@@ -0,0 +1,21 @@
...
+        return None
+
+    fib = np.zeros(u, dtype=np.int64)  # Initialize array for Fibonacci numbers
+    fib[1] = 1  # Set the first Fibonacci number
+
+    for i in range(2, u):
+        fib[i] = fib[i - 1] + fib[i - 2]
+
+    # Sum the Fibonacci numbers
+    r = np.sum(fib[1:])  # Exclude the 0th Fibonacci number for consistency
```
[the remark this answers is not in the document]
Karatsidis: sounds good; I have updated the implementation to maintain the recursion.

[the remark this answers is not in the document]
axman6: You haven't really, at best this is just obfuscating that this is using the linear time algorithm. The point of this benchmark is to show the performance of the naive recursive implementation of fibonacci. Otherwise we'd use the classic Haskell `fibs = 1:1:zipwith (+) fibs (fail fibs); fib n = fibs !! n` which does the same thing that's happening in your code. Your code will make `N` calls to `fibonacci`, not `fibonacci(N)` calls as is expected. My understanding is that in all languages, their implementation of fibonacci should have `O(fibonacci(N))` runtime as written (if the compiler is able to optimise that into something better, then that's a feature of the language that it can perform those optimisations without the developer having to rewrite their code).

fibonacci/numpy/code.py
```diff
@@ -0,0 +1,36 @@
+import sys
+import numpy as np
+
+
+def fibonacci(n, fib_array):
+    # If the Fibonacci number has already been calculated, return it
+    if fib_array[n] != -1:
+        return fib_array[n]
```
bddicken, line 8: This is basically a cache for the results? We could do the same in the other languages and also get a huge boost, so this is not going to be accepted.

PR #144, opened 2024-11-29
Add OCaml benchmark

juliojgd:


fibonacci/ocaml/code.ml
```diff
@@ -0,0 +1,17 @@
+
+
+let fibonacci n =
+  let rec aux a b n =
+    if n = 0 then a
+    else aux b (a + b) (n - 1)
+  in
```
bddicken, line 7: I don't know Ocaml. Does this match the other naive implementation from the other languages?

botantony: I do not use functional languages that much but this program looks exactly like if someone asked me to implement Fibonacci series with naive approach. The only difference is that we go from bottom to top but I cannot imagine how to do it the other way using functional paradigm

juliojgd: This is a recursive implementation the same as the other but with a difference. It's tail call recursion. So OCaml compiler can optimize in an extreme way such type or recursion. 

So, it's recursion, no caching is done, and the good performance comes from the use of tail calls.

But for the sake of comparison, I'll change it to "non tail call"recursion.

juliojgd: Changed. It's important to note that tail call recursion is always preferred (I guess in any language) in OCaml.

botantony (approved):
it works

SGrondin:
Just adding my 2¢ in passing, but using tail-recursion (TCO) in OCaml is something taught and emphasized from the very beginning.

The current code (with TCO removed) is simply never done, it's not an accurate representation of real OCaml code. The original version was correct, yes, we modify **every** recursive algorithm to make them tail-recursive whenever possible. That's just how we do it.

PEZ:
I think this looks compliant, but it is outdated and needs to be rewritten to target the current benchmark runner. See the README about this, and also #371.

PR #146, opened 2024-11-29
Optimizing PHP loops benchmark by using reference operator

MuriloChianfa:
Hey guys, I've tested this new implementation for PHP and I'm getting a little speedup by using the [PHP reference operator](https://www.php.net/manual/en/language.references.php).

The concept here is because we already have the entire array initialized by the array_fill function, so why not access each position by their reference to make each change without need to re-access the same $a[$i] value so much times.

PS.: It's possible to do with "for" instead "foreach" too, creating a new temp variable by de-referencing the $a[$i] value with something like $c =& $a[$i], but I used "foreach" because I know that have some discussions around them 😂

Please test my changes to see if it's a really speedup!

## Before:
php-benchmark-loops-before-optimization

## After:
php-benchmark-loops-after-optimization

### DIFF
--- a/loops/php/code.php
+++ b/loops/php/code.php
@@ -3,10 +3,10 @@
 $u = (int) $argv[1];                                # Get an input number from the command line
 $r = rand(0, 10000);                                # Get a random number 0 <= r < 10k
 $a = array_fill(0, 10000, 0);    # Array of 10k elements initialized to 0
-for ($i = 0; $i < 10000; $i++) {                    # 10k outer loop iterations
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for the contribution. I'm a little hesitant to merge a change like this, as in some sense it gives an unfair advantage since it now gets to skip a bunch of array index operations. There's other languages that would benefit from changes like this if I allowed them too....

MuriloChianfa:
Let's do both scenarios and compare witch language have better results for all possible scenarios and code implementations, perhaps is a good idea but will provide more work LOL

PR #148, opened 2024-11-30
Add vectorized version of java loop

tjake:
This pull request includes several changes to improve the performance and functionality of the Java code by utilizing the `jdk.incubator.vector` module. Additionally, there are minor adjustments to the `run.sh` script to accommodate these changes.

With these changes java impl is ~1.8x faster than c on my machine.
```
C = 1.21users
C = 1.22users
C = 1.22users

Go = 1.24users
Go = 1.24users
Go = 1.22users

Rust = 1.21users
Rust = 1.21users
Rust = 1.22users

Java = 0.78users
Java = 0.76users
Java = 0.76users
```

### Enhancements to Java code:

* [`compile.sh`](diffhunk://#diff-24d024fa911cc01760d679370d8ea8711230f073fde3e721a59e292b825252c4L3-R3): Updated the `javac` command to include the `jdk.incubator.vector` module for vector API support.
* [`loops/jvm/code.java`](diffhunk://#diff-d2611cfdd8159d2ed1167c1de75cc7b11aed3cb01d8981de431cc87056d539f4L5-R21): Refactored the code to use the vector API for improved performance. This includes broadcasting scalars to vectors, vectorized addition, and storing vectors back to arrays.

### Adjustments to run script:

* [`run.sh`](diffhunk://#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35dL4-R4): Modified the `runOnce` function to filter for the `user` time in the output, this was required because java includes a incubator warning to stderr.
* [`run.sh`](diffhunk://#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35dL23-R23): Updated the Java run command to include the `jdk.incubator.vector` module.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,6 +1,6 @@
 clang -O3 c/code.c -o c/code
 go build -ldflags "-s -w" -o go/code go/code.go
-javac jvm/code.java
+javac --add-modules jdk.incubator.vector jvm/code.java
 cargo build --manifest-path rust/Cargo.toml --release
 kotlinc -include-runtime kotlin/code.kt -d kotlin/code.jar
 #kotlinc-native -include-runtime kotlin/code.kt -d kotlin/code
--- a/loops/jvm/code.java
+++ b/loops/jvm/code.java
@@ -2,17 +2,23 @@ package jvm;
 
 import java.util.Random;
 
-public class code {
+import jdk.incubator.vector.IntVector;
+import jdk.incubator.vector.VectorOperators;
 
+public class code {
     public static void main(String[] args) {
         var u = Integer.parseInt(args[0]); // Get an input number from the command line
         var r = new Random().nextInt(10000); // Get a random number 0 <= r < 10k
+        var rv = IntVector.SPECIES_PREFERRED.broadcast(r); // Broadcast scalar to vector
         var a = new int[10000]; // Array of 10k elements initialized to 0
-        for (var i = 0; i < 10000; i++) { // 10k outer loop iterations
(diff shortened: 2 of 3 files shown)


bddicken:
If this is 1.8x faster than C and Rust, I'm very sceptical. Not that the timings are a lie, but that this is giving java some unfair advantage by allowing it to do less work.

I don't know much about `jdk.incubator.vector`. Can you elaborate on exactly what is going on under the hood here?

The idea with this benchmark is not necessarily to what a "faster" version of the program, but to write the "same" code in many languages, and see which perform the best based on compiler / interpreter / jit optimizations and such.

tjake:
> Can you elaborate on exactly what is going on under the hood here?

Sure, this is just SIMD which Java now natively supports arm and intel. 

No problem if this is outside the goal of your experiment. I thought it was fair game because there’s a c simd PR you seemed excited by
https://github.com/bddicken/languages/pull/48

neon-sunset:
@bddicken This implementation vectorizes the code explicitly. However, it performs the work differently. The same can be done for other languages (Panama Vectors in Java are still in experimental mode and are not production-ready, but other languages do expose platform or portable SIMD APIs of similar type). However, it's not that different from just pre-computing the value in the shortest manner possible which will not perform the baseline work. It's not dissimilar to explicitly parallelizing the code except at instruction level instead.

bddicken:
OK, so this is SIMD. In that case, though interesting, I am not going to include this in the benchmark.

PR #151, opened 2024-12-01
Added Scala; fixed some awkward things in scripts

vpatryshev:
The biggest piece is added `setup.sh`.
Also, `run.sh` checks for the presence of a language before launching.

And Scala3 added. Scala3 seems to be slower than expected. And more awkward to configure. 
Each Scala project requires 4-5 files to be added.

Another thing: `hyperfine` is available on linux, but not on mac. So I kept the old solution for the case it's not there.


bddicken:
Thanks for the submission. Scala was added via a different PR. I'm going to close this one out, but feel free to check out the existing implementations and make a PR if you have suggestions for improving.

PR #153, opened 2024-12-02
Implement vectorized and mathematically optimized modulo sum for 1 bi…

AfolabiOlaoluwa:
…llion iterations

The original solution performed a nested loop with 10,000 outer iterations and 100,000 inner iterations, resulting in a total of 1 billion modulo operations. This approach, while straightforward, is computationally expensive and lacks optimization.

To address this, the optimized solution has been implemented with the following key improvements:

1. Vectorization:
   - The new method uses the Array.new construct with a block to generate the result array in a single pass, avoiding explicit nested loops.
   - This vectorized approach eliminates the need for manual loop management and improves overall efficiency.

2. Mathematical Optimization:
   - Instead of relying on the iterative modulo computations, the new method uses a mathematical formula to approximate the sum of the modulo operations.
   - The formula `(u * (u - 1)) / 2 * (100_000 / u) + (100_000 % u) * ((100_000 % u) - 1) / 2` captures the essence of the modulo summation in a more efficient manner.

These optimizations result in a dramatic performance improvement, reducing the 'total' time from 73.545019 to 0.001094, a speed-up of over 67,000x.

The commit ensures the optimized solution provides the same computational result as the original nested loop approach, while dramatically improving the execution speed for 1 billion iterations through the use of vectorization and mathematical optimization.

**Benchmark check code**:

```ruby
def original_method(u)
  r = rand(10_000)
  a = Array.new(10_000, 0)

  10_000.times do |i|
    100_000.times do |j|
      a[i] += j % u
    end
    a[i] += r
  end

  a[r]
end

def optimized_method(u)
  r = rand(10_000)

  a = Array.new(10_000) do
    sum = (u * (u - 1)) / 2 * (100_000 / u) + (100_000 % u) * ((100_000 % u) - 1) / 2
    sum + r
  end

  a[r]
end

# Verification
[1, 2, 3, 4, 5, 10].each do |u|
  srand(42)
  original_result = original_method(u)

  srand(42)
  optimized_result = optimized_method(u)

  puts "u = #{u}"
  puts "Original method result: #{original_result}"
  puts "Optimized method result: #{optimized_result}"
  puts "Difference: #{original_result - optimized_result}"
  puts "---"
end

# Performance Benchmark
require 'benchmark'

Benchmark.bm(20) do |x|
  x.report("Original Method:") { original_method(5) }
  x.report("Optimized Method:") { optimized_method(5) }
end


#-----
def exact_sum(u)
  (0...100_000).sum { |j| j % u }
end

def analytical_attempt(u)
  (u * (u - 1)) / 2 * (100_000 / u) + (100_000 % u) * ((100_000 % u) - 1) / 2
end

[1, 2, 3, 4, 5, 10].each do |u|
  puts "u = #{u}"
  puts "Exact sum:      #{exact_sum(u)}"
  puts "Analytical sum: #{analytical_attempt(u)}"
  puts "Difference:     #{exact_sum(u) - analytical_attempt(u)}"
  puts "---"
end
```

**Benchmark result**:
Screenshot 2024-12-02 at 14 39 54


artemisart:
The point of the benchmark is to do these 10k*100k iterations and all modulo operations, this is not a "lack of optimization". This is the same for the fibonacci implementations which are an easy way to benchmark recursion and function calling but not efficient to compute fibonacci numbers.

AfolabiOlaoluwa:
> The point of the benchmark is to do these 10k*100k iterations and all modulo operations, this is not a "lack of optimization". This is the same for the fibonacci implementations which are an easy way to benchmark recursion and function calling but not efficient to compute fibonacci numbers.

Interesting! I did ask that question before working on it. That's okay. It can be closed.

bddicken:
Thanks for submitting this @AfolabiOlaoluwa. As @artemisart pointed out, the point is not to submit the most mathematically-optimized possible solution, but have approximately-equivalent solutions in terms of code flow and computations in each language. Yes, this is somewhat of a silly program, but the purpose is for performance comparison.

PR #156, opened 2024-12-03
Cheating-less, 1.5 times faster Python implementation of the Fibonacci benchmark

FI-Mihej:
This PR provides the following improvements:

- Adopted the ternary operator as a fully common, recommended, and Pythonic style.
- Reflects the existing approaches (including ternary-operator-based approach) used in other implementations:
    - C#: `n <= 1 ? n : Fibonacci(n - 1) + Fibonacci(n - 2);` - ([latest code.cs](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/csharp/code.cs))
    - Inko: `if n < 2 { n } else { fibonacci(n - 1) + fibonacci(n - 2) }` - ([latest code.inko](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/inko/code.inko))
    - Elixir: `def fib(n) when n <= 1, do: n` - ([latest bench.exs](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/elixir/bench.exs))
    - Nim: `if n <= 1:` - ([latest code.nim](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/nim/code.nim))
- **CPython**: Significantly faster than the current Python implementation - **around 1.2 to 1.5 times faster**, depending on the CPython interpreter version.
- **PyPy**: Slightly faster than the current Python implementation - around 1.03 to 1.06 times faster, depending on the PyPy interpreter version.

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -1,11 +1,7 @@
 import sys
 
 def fibonacci(n):
-  if (n == 0):
-    return 0
-  if (n == 1):
-    return 1
-  return fibonacci(n-1) + fibonacci(n-2)
+  return n if n <= 1 else fibonacci(n-1) + fibonacci(n-2)
 
 def main():
     u = int(sys.argv[1])


FI-Mihej:
Additionally, the following implementations use "pattern-matching-based" or at least "case-based" approaches, which are entirely optional for their respective languages but significantly faster:

- **Crystal**: [latest code.cr](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/crystal/code.cr)
- **Julia**: [latest code.jl](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/julia/code.jl)
- **Rust**: [latest code.rs](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/rust/code.rs)
- **Scala**: [latest code.scala](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/scala/code.scala)
- **Zig**: [latest code.zig](https://github.com/bddicken/languages/blob/bf71f98f0a823775c16fdb79f2445438f471edf6/fibonacci/zig/code.zig)

bddicken:
Thanks for submitting this! It's cool that this gives a little performance boost. However, I think the solution here is it either needs to be changed in *all* the languages to use this style solution, or the others need to be updated. I think there are less of the `<= 1` style at the moment, so I'll probably swing the other way.

FI-Mihej:
> Thanks for submitting this! It's cool that this gives a little performance boost. However, I think the solution here is it either needs to be changed in _all_ the languages to use this style solution, or the others need to be updated. I think there are less of the `<= 1` style at the moment, so I'll probably swing the other way.

Oh, come on. We all get it. You could’ve just ignored it or quietly closed the PR without any explanation - no big deal. It’s normal to have emotions (personal likes and irrational dislikes) towards tools - like a pair of pliers - it happens.
A few years ago, Erlang went through something similar: hundreds of developers on blogs and Twitter absolutely hated on it for no objective reason. Nothing to be ashamed of :D

axman6:
@FI-Mihej This is a really disrespectful way to respond to some very mild negative news. I personally think your changes here are fine (and would also be happy if all implementations changed to using `return 1 if n <= 1` because it doesn't fundamentally change the algorithm being tested (maybe there's a constant factor improvement but not an algorithmic one).

Just remember, this is one person's fun project they're working on in their own time, they don't need people getting angry at them when they owe you nothing.

FI-Mihej:
> @FI-Mihej This is a really disrespectful way to respond to some very mild negative news. I personally think your changes here are fine (and would also be happy if all implementations changed to using `return 1 if n <= 1` because it doesn't fundamentally change the algorithm being tested (maybe there's a constant factor improvement but not an algorithmic one).
> 
> Just remember, this is one person's fun project they're working on in their own time, they don't need people getting angry at them when they owe you nothing.

Disrespect is deliberate gaslighting - when someone, in specific cases, repeatedly hides behind fabricated rationalizations of their emotional bias towards tools: some tools are allowed everything, while others aren't allowed anything.

If this behavior only affects an obscure project, it doesn't really matter. But when such biased opinions are spread to the masses, it becomes large-scale gaslighting.

Spotting this trend isn't difficult: just spend an hour reviewing all the accepted and rejected pull requests.

PR #158, opened 2024-12-03
Fish implementations for loop & fibonacci

ThatOneCalculator:
Implementation for https://fishshell.com/, tested against v3.7.1


bddicken:
Thanks for submitting this! It needs a little update before we can merge.

ThatOneCalculator:
@bddicken should be all fixed now!

PR #159, opened 2024-12-03
[D] Arm M1

cyrusmsk:
For usage with M1 Arm is better to add this compilation flag.


cyrusmsk:
To run code you need to first give an approve to the LTO lib. To do that on mac you need just to open in the terminal file : "d_path"/lib/libLTO.dylib
After that the compilation will be possible

bddicken:
Nice, thanks.

PR #160, opened 2024-12-03
Cleanup pascal code

turric4n:


turric4n:
Improve Pascal performance.

fibonacci/fpc/code.pas
```diff
@@ -3,26 +3,15 @@
 uses
   SysUtils;
 
-function Fibonacci(n: Int32): Int32;
+function Fibonacci(n: UInt32): UInt32; inline;
```
bddicken, line 6: What does `inline` do here?

fibonacci/fpc/code.pas
```diff
@@ -3,26 +3,15 @@
...
 begin
-  if n = 0 then
-    Fibonacci := 0
-  else if n = 1 then
-    Fibonacci := 1
-  else
-    Fibonacci := Fibonacci(n - 1) + Fibonacci(n - 2);
+  if n <= 1 then exit(n);
```
bddicken, line 8: Let's revert this back to the if/else.

loops/fpc/code.pas
```diff
@@ -4,28 +4,20 @@
   SysUtils;
 
 var
-  u, r, i, j: Int32;
-  a: array[0..9999] of Int32;
+  u: UInt32;
+  a: array[0..9999] of UInt32;
```
bddicken, line 8: Interesting! Does using a unit give better performance than a signed iint?

turric4n: Surprisingly, yes, it's 40% more efficient.

fibonacci/fpc/code.pas
turric4n: Procedures that are declared inline are copied to the places where they are called. This has the effect that there is no actual procedure call, the code of the procedure is just copied to where the procedure is needed, this results in faster execution speed if the function or procedure is used a lot. It is obvious that inlining large functions does not make sense.

fibonacci/fpc/code.pas
turric4n: ok,  no problem

bddicken:
Nice, thanks!

PR #161, opened 2024-12-03
Replace NSMutableArray with CFMutableArray

denizak:


bddicken:
Thanks for submitting this. I's been a long time since I written Objective-C, so I need some context here. What is the difference between the NS and CF variants? Why is this change better for performance? Have some timings to back it up?

Thanks.

denizak:
attached runtime test
Screenshot 2024-12-08 at 13 14 58

bddicken:
OK, but I still need further explanation of what the change is actually doing, please!

denizak:
This PR is about changing the usage of NSMutableArray into CFMutableArray. Using the non-retaining CFMutableArray to make the process faster and reduce the retain time of each object being append to the array.

vendethiel:
The docs seem to indicate that they are the same underlying type

PEZ:
Closing, because the array has been changed to be a stack-allocated C-style array by another change.

PR #162, opened 2024-12-04
Add  Clojure and Babashka (both loops and fibonacci)

PEZ:
Hi, thanks for creating this repo and those cool video graphs!

Here are implementations for both loops and fibonacci in my language of choice: Clojure.

The Clojure CLI tools are needed: https://clojure.org/guides/install_clojure

Please don't hesitate to let me know if I need to fix something.

**UPDATE**: Also adds my _scripting_ language of choice: [Babashka](https://babashka.org/). It runs on the same source as the Clojure entry does (for now, at least). But you'll need the `bb` command line tool to run it. https://github.com/babashka/babashka#installation


ghoseb:
@PEZ Thanks! Is it possible to avoid the `clojure -Spath` call for each run? That classpath finding code could be really slow and will affect each run. Perhaps we could compile an uberjar instead? Also, is it necessary to implement fibonacci using an interface/deftype? We could consider sticking to idiomatic Clojure (unless the speedup is 100x or something of that order). Let me know what you think.

PEZ:
@ghoseb Thanks for the feedback! I think the expansion of the path happens before it is passed to the benchmark runner (hyperfine). The run output looks like so:

```
Benchmarking C
Benchmark 1: ./c/code 40
  Time (mean ± σ):     247.6 ms ±   1.0 ms    [User: 244.5 ms, System: 2.3 ms]
  Range (min … max):   246.4 ms … 248.3 ms    3 runs
 

Benchmarking Java
Benchmark 1: java jvm.code 40
  Time (mean ± σ):     355.8 ms ±  10.7 ms    [User: 329.6 ms, System: 23.0 ms]
  Range (min … max):   344.3 ms … 365.6 ms    3 runs
 

Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 40
  Time (mean ± σ):     548.8 ms ±   9.9 ms    [User: 789.6 ms, System: 59.5 ms]
  Range (min … max):   538.7 ms … 558.5 ms    3 runs
```

The interface was not gaining any 100X. More like 25%. But I can use an anonymous function instead, avoiding the function lookup cost. Like so:

``` clojure
(defn -main [& args]
  (let [fibonacci (fn fib ^long [^long n]
                    (if (<= n 1)
                      n
                      (+ (long (fib (- n 1)))
                         (long (fib (- n 2))))))
        ^long u (parse-long (first args))
        ^long r (loop [i 1
                       sum 0]
                  (if (< i u)
                    (recur (inc i) (+ sum (fibonacci i)))
                    sum))]
    (println r)))
```

But maybe I can also give the compiler some option to make the function call less dynamic...

ghoseb:
Thank you @PEZ I really like the anonymous function solution. If that's not too much of a slowdown compared to the interface, please feel free to modify the PR. Cheers!

ghoseb (approved):
Looks good to me.

PEZ:
It's the same performance with the anonymous function as with the interface + deftype shenanigans.

ghoseb:
@PEZ Just a thought, are there any JVM flags that we can use that causes a shorter startup time? I remember `lein` used certain flags like `-XX:+TieredCompilation -XX:TieredStopAtLevel=1` to make the launch time faster. Since this is not a long running process, not sure if it's worth running the JIT etc at all.

PEZ:
I'm game for anything that can get the JVM start to play less of a part. Just throwing those options you had there in made the benchmark runner spend a lot of time in warmup and seems to have hurt a lot:

```
Benchmarking Clojure
Benchmark 1: java  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 40
  Time (mean ± σ):      5.120 s ±  0.086 s    [User: 5.100 s, System: 0.066 s]
  Range (min … max):    5.032 s …  5.203 s    3 runs
```

I really don't understand what I just tried. 😄

compile.sh
```diff
@@ -27,3 +27,4 @@ swiftc -O -parse-as-library -Xcc -funroll-loops -Xcc -march=native -Xcc -ftree-v
 #dotnet publish csharp/csharp.csproj -o csharp/code-aot /p:PublishAot=true
 dotnet publish csharp/csharp.csproj -o csharp/code
 ghc -O2 -fllvm haskell/code.hs -o haskell/code || { echo "ghc: cannot compile with llvm backend; fallback to use default backend"; ghc -O2 haskell/code.hs -o haskell/code; }
+cd clojure && mkdir -p classes && clojure -Sdeps '{:paths ["."]}' -M -e "(compile 'code)"
```
N-litened, line 30: I think it's a good idea to restore the working directory back after we `cd`, so that other languages added after this don't need to track the current directory changes 🙏

PEZ: Good catch. I actually ran in to this with the `inko` compile, since I don't have it, it crashed out before it could `cd` back up again and the rest of the compiles failed. I got around that by executing the `inko` commands in a sub shell. Added that to the Clojure compile now. WDYT?

PEZ:
I made the `fibonacci` function iterate instead of recurse, and that seems to have helped the compiler a lot. And then it doesn't matter if I use a namespace function instead of a local one. Benchmark output:

```
Benchmarking C
Benchmark 1: ./c/code 40
  Time (mean ± σ):     256.3 ms ±   2.1 ms    [User: 253.2 ms, System: 2.5 ms]
  Range (min … max):   254.7 ms … 258.7 ms    3 runs
 

Benchmarking Java
Benchmark 1: java jvm.code 40
  Time (mean ± σ):     356.3 ms ±  11.0 ms    [User: 328.0 ms, System: 24.0 ms]
  Range (min … max):   347.8 ms … 368.7 ms    3 runs
 

Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 40
  Time (mean ± σ):     279.4 ms ±  11.6 ms    [User: 428.8 ms, System: 46.5 ms]
  Range (min … max):   271.1 ms … 292.6 ms    3 runs
```

So, almost twice as fast as with the recursion, and even faster than Java. On my machine, at least.

PEZ:
Hmmm, maybe not:

> **fibonacci**

I'll revert that change.

lost22git:
What about run with babashka or without gc `java -XX:+UseEpsilonGC` :P

PEZ:
We could add Babashka with this PR too, I guess:

```
Benchmarking Babashka
Benchmark 1: bb -cp clojure -m code 40
  Time (mean ± σ):     35.688 s ±  0.076 s    [User: 35.488 s, System: 0.130 s]
  Range (min … max):   35.619 s … 35.770 s    3 runs
```

Even if it may be that we can write a faster bb fibbonaci than this one. 😄

PEZ:
I don't think the GC ever runs for these programs. But we can make the JVM and Clojure start penalties less significant by running e.g. fibonacci to, say, 48:

```
❯ bash ../run.sh 48

Benchmarking C
Benchmark 1: ./c/code 48
  Time (mean ± σ):     12.109 s ±  0.053 s    [User: 12.053 s, System: 0.039 s]
  Range (min … max):   12.053 s … 12.159 s    3 runs
 

Benchmarking Java
Benchmark 1: java jvm.code 48
  Time (mean ± σ):     14.891 s ±  0.074 s    [User: 14.797 s, System: 0.073 s]
  Range (min … max):   14.842 s … 14.976 s    3 runs
 

Benchmarking Clojure
Benchmark 1: java -XX:+UseZGC -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 48
  Time (mean ± σ):     14.253 s ±  0.179 s    [User: 14.367 s, System: 0.095 s]
  Range (min … max):   14.073 s … 14.430 s    3 runs
```

That seems to say Clojure runs something like 17% slower for this test? I would never have guessed.

PEZ:
And now I am not too sure about the GC thing. I noticed that in the run above I was still using ZGC from an earlier experiment.

Running without specifying a GC I get a bit different results:

```
Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 48
  Time (mean ± σ):     12.839 s ±  0.056 s    [User: 12.970 s, System: 0.091 s]
  Range (min … max):   12.797 s … 12.903 s    3 runs
```

Which looks very good! But in a second run I got:

```
Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.11.2/clojure-1.11.2.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar code 48
  Time (mean ± σ):     19.486 s ± 11.172 s    [User: 19.572 s, System: 0.132 s]
  Range (min … max):   13.017 s … 32.387 s    3 runs
```

Look at the *Range*! Not sure if the GC _can_ have that impact. But anyway.

I take home that Clojure can run at only 6% slower than C. 😄

bddicken:
Awesome, thanks for the detailed discussion on this 👍.

PEZ:
Thanks for providing this project, @bddicken! 🙏 ❤️

PEZ:
I added an issue about the start time thing here:
* #220

PR #163, opened 2024-12-04
Remove duplicated entries and sort

Gutem:
pypy and ruby was duplicated.
the run file was sorted


bddicken:
Thanks for this. I already removed the duplicates, and I actually like having more control than alpha over the order of the languages running.

PR #164, opened 2024-12-04
Add bash (loop only)

molocohrrr:
Adding bash language, only loops implementation


molocohrrr:
I will re-open with fibonacci bench included :-)

PR #165, opened 2024-12-04
chore: optimize scala performance

He-Pin:
IIRC, Scala 2.13 does the trick where it transform the for  to while loop, have not checked the Scala 3.6.x
docs: https://scala-cli.virtuslab.org/docs/reference/directives

### DIFF
--- a/fibonacci/scala/code.scala
+++ b/fibonacci/scala/code.scala
@@ -1,16 +1,21 @@
+//> using scala 3.6.1
+
 package jvm
+
+import scala.annotation.switch
 
 @main
 def main(number: String): Unit =
   val u = number.toInt
   var r = 0
-  for i <- 1 until u do
+  var i = 0
+  while i < u do
     r += fibonacci(i)
+    i += 1
   println(r)
 
-
 def fibonacci(n: Int): Int =
-  n match
+  (n: @switch) match
     case 0 => 0
     case 1 => 1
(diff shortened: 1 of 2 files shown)


He-Pin:
@WojciechMazur Does Scala 3 transform the `for` into a `while loop`?
It's slower than Java currently

He-Pin:
@bddicken If naive fib is required I can paritally revert it.

fibonacci/scala/code.scala
```diff
@@ -1,16 +1,24 @@
...
+    r += fibonacci(i, 0, 1)
+    i += 1
   println(r)
 
+@tailrec
+def fibonacci(n: Int, a: Int, b: Int): Int =
+  if (n > 0)
+    fibonacci(n - 1, b, a + b)
+  else
+    a
```
axman6, line 22: As far as I understand it, the point of this benchmark is to test the naive non-tail recursive implementation. Every language would be faster if it used the linear time algorithm.

He-Pin: seems true, let's me revert it

He-Pin:
Currently , Range.foreach is not optimized to while loop
https://github.com/scala/scala3/issues/8880
https://github.com/scala/bug/issues/1338

He-Pin:
@bddicken Please merge this , thanks.

bddicken:
Thanks for submitting this 👍

PR #166, opened 2024-12-04
Add Lean 4 loops program

RichardKelley:
Added an implementation of the loops benchmark in the Lean 4 programming language and theorem prover. On my M3 Macbook the benchmark comes in around 1.6 seconds.

Builds with the lake build system, which can be installed using elan, the lean4 version manager.


bddicken:
Thanks for submitting this!

Could you add a fibonacci solution as well?

RichardKelley:
Fibonacci added. Runs in about 550ms.

bddicken:
👍

PR #167, opened 2024-12-04
Add Numpy; Mac Support; Minor fixes

cpeters008:
add numpy; fix haskell; add mac os brew install;
add gnu parallel run script; update clean; minor fixes to compile


TurnrDev:
Using Numpy on a Python comparison is cheating as it calls into C code

loops/.gitignore
axman6, line 1: Does a) this .gitignore need to exist, and b) does it need to be executable???

fibonacci/cpp/code.cpp
```diff
@@ -0,0 +1,30 @@
...
+    if (n == 0) return 0;
+    if (n == 1) return 1;
+    int32_t a = 0, b = 1, c;
+    for (int32_t i = 2; i <= n; i++) {
+        c = a + b;
+        a = b;
+        b = c;
+    }
+    return b;
+}
```
axman6, line 15: This doesn't use the recursive algorithm, which AFAIUI, is the point of this benchmark.

compile.sh
```diff
@@ -26,4 +34,8 @@ swiftc -O -parse-as-library -Xcc -funroll-loops -Xcc -march=native -Xcc -ftree-v
 # haxe --class-path haxe -main Code --jvm haxe/code.jar # was getting errors running `haxelib install hxjava`
 #dotnet publish csharp/csharp.csproj -o csharp/code-aot /p:PublishAot=true
 dotnet publish csharp/csharp.csproj -o csharp/code
-ghc -O2 -fllvm haskell/code.hs -o haskell/code || { echo "ghc: cannot compile with llvm backend; fallback to use default backend"; ghc -O2 haskell/code.hs -o haskell/code; }
+cabal update
+cabal install --lib random --force
+ghc -O2 -fllvm -Wno-partial-fields haskell/code.hs -o haskell/code || { echo "ghc: cannot compile with llvm backend; fallback to use default backend"; ghc -O2 -Wno-partial-fields haskell/code.hs -o haskell/code; }
```
axman6, line 39: If we're using llvm, we should also tell it to optimise its results.

```suggestion
ghc -O2 -fllvm -optlo -O3 -Wno-partial-fields haskell/code.hs -o haskell/code || { echo "ghc: cannot compile with llvm backend; fallback to use default backend"; ghc -O2 -Wno-partial-fields haskell/code.hs -o haskell/code; }
```

bddicken:
Thanks for submitting this. It seems like there are some very cool improvements here (for example, having `brew_mac_install.sh` will be nice!). However, there's also a bunch of excessive changes here. We need to keep each PR so a smaller set of changes.

Would you be willing to split these up into separate PRs for each change? ?If not, I can later come through and cherry-pick just the things I want to bring in.

bddicken:
Closing this one out, too many things going on at once.

PR #168, opened 2024-12-04
improve golang

smallnest:
use a temp to avoid accessing a[i] each loop. and use gofmt to format code

### DIFF
--- a/loops/go/code.go
+++ b/loops/go/code.go
@@ -1,22 +1,26 @@
 package main
+
 import (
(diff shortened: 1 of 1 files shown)


smallnest:
image

komuw:
also see(and conversation there in); https://github.com/bddicken/languages/pull/150

smallnest:
GOOD.  duplicated. I close this.

PR #169, opened 2024-12-04
Use stackalloc in C#

Golle:
Use stackalloc instead of new.

Would only be fair since that's what the C sample does!  🤓 

Is it faster? probably a few nanonseconds.


bddicken:
cool!

PR #172, opened 2024-12-04
add awk/mawk fibonacci & loops

taprs:
Tested with GNU Awk v5.2.1 and mawk 1.3.4 20240123, should be good with any POSIX AWK including nawk.


bddicken:
this is amazing 👍

PR #175, opened 2024-12-04
R-appropriate implementation

aadler:
Your use of nested loops in R is not a proper comparison, as R is meant to be used as a vectorized language. The attached `code2.R` implements a vectorized version of your double loop and is what should be programmed to solve the problem in the R ecosystem. I'd request that the timing of this version be shown as well.

### DIFF
--- a/loops/r/code2.R
+++ b/loops/r/code2.R
@@ -6,7 +6,7 @@ u <- as.integer(args[1])  # Convert to integer
 r <-  sample.int(10000, 1) - 1L
 
 # Vectored version of loop
-a <- rep(sum((seq_len(100000) - 1L) %% u), 10000) + r
+a <- rep(sum(0:99999 %% u), 10000) + r
 
 # Print a single element from the array
 print(a[r + 1])  # Adjusted for R's 1-based indexing
--- a/loops/r/code2.R
+++ b/loops/r/code2.R
@@ -1,0 +1,12 @@
+# Get input number from the command line arguments
+args <- commandArgs(trailingOnly = TRUE)
+u <- as.integer(args[1])  # Convert to integer
+
+# Generate a random number between 0 and 10000
+r <-  sample.int(10000, 1) - 1L
+
+# Vectored version of loop
+a <- rep(sum((seq_len(100000) - 1L) %% u), 10000) + r
+
+# Print a single element from the array
+print(a[r + 1])  # Adjusted for R's 1-based indexing
--- a/loops/r/code2.R
+++ b/loops/r/code2.R
@@ -2,7 +2,7 @@
 args <- commandArgs(trailingOnly = TRUE)
 u <- as.integer(args[1])  # Convert to integer
 
-# Generate a random number between 0 and 10000
+# Generate a random number between 0 and 9999
 r <-  sample.int(10000, 1) - 1L
 
 # Vectored version of loop


loops/r/code2.R
```diff
@@ -0,0 +1,12 @@
+# Get input number from the command line arguments
+args <- commandArgs(trailingOnly = TRUE)
+u <- as.integer(args[1])  # Convert to integer
+
+# Generate a random number between 0 and 10000
+r <-  sample.int(10000, 1) - 1L
```
J-Moravec, line 6: `r <- sample.int(0:9999, 1)` might be a tiny bit faster with ALTREP expansion done in the C code.

aadler: Hi, @J-Moravec. That's a good idea. However, unlike `sample`, `sample.int` does not accept a vector from which to sample. Rather, it accepts a number which represents the magnitude of the set from which to sample. Therefore, your suggestion won't work. Using `sample.int` with the broadcast subtraction is still faster than using sample, since we know the magnitude of the set, as shown below:

```r
library(microbenchmark)
microbenchmark(sample.int(10000, 1) - 1L, sample(0:9999, 1), check = "identical", control = list(order = "block"), setup = set.seed(54), times = 1e5L)

Unit: microseconds
                      expr min  lq     mean median  uq    max neval cld
 sample.int(10000, 1) - 1L 3.7 4.1 4.465435    4.2 4.3 2831.0 1e+05  a 
         sample(0:9999, 1) 4.7 5.2 5.895089    5.4 5.7 5786.1 1e+05   b

sessionInfo()
R Under development (unstable) (2024-11-21 r87359 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 10 x64 (build 19045)

Matrix products: default


locale:
[1] LC_COLLATE=English_United States.utf8  LC_CTYPE=English_United States.utf8    LC_MONETARY=English_United States.utf8
[4] LC_NUMERIC=C                           LC_TIME=English_United States.utf8    

time zone: Asia/Jerusalem
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] microbenchmark_1.5.0

loaded via a namespace (and not attached):
 [1] MASS_7.3-61      TH.data_1.1-2    zoo_1.8-12       compiler_4.5.0   Matrix_1.7-1     multcomp_1.4-26 
 [7] sandwich_3.1-1   tools_4.5.0      survival_3.7-0   mvtnorm_1.3-2    codetools_0.2-20 splines_4.5.0   
[13] grid_4.5.0       lattice_0.22-6 
```

aadler: However, in the actual calculation, using the explicit sequence generation via the colon operator IS faster, and is safe given we know the loop boundaries already. See commit 78da1e39ca8891953af014adb7aba83676a5cd7f.

````r
microbenchmark(seq_len(100000) - 1L, 0:99999, check = "identical", control = list(order = "block"), times = 1e5)
Unit: nanoseconds
                expr    min     lq       mean median     uq       max neval cld
 seq_len(1e+05) - 1L 161400 168200 272171.163 171400 201100 103369400 1e+05  a 
             0:99999      0      0     96.752      0    100   4852900 1e+05   b
```

J-Moravec: Thanks for correcting me. I should have benchmarked it before writing. I did some benchmarking a week ago when this story first landed, but turns out I did not remember stuff exactly.

In fact, my suggested solution is not even consistently faster, and is often slower than the `sample(10000, 1) - 1`.

loops/r/code2.R
```diff
@@ -0,0 +1,12 @@
+# Get input number from the command line arguments
+args <- commandArgs(trailingOnly = TRUE)
+u <- as.integer(args[1])  # Convert to integer
+
+# Generate a random number between 0 and 9999
+r <-  sample.int(10000, 1) - 1L
+
+# Vectored version of loop
+a <- rep(sum(0:99999 %% u), 10000) + r
```
axman6, line 9: I assume that `sum(0:99999 %% u)` will only be calculated once here? that seems to make this fundamentally different to the other implementations which perform all the calculations multiple times (the is the reason I [suggested that the Haskell](https://github.com/bddicken/languages/pull/167#discussion_r1870507602) one be forced to do all requested work - due to haskell's laziness, only the `r`'th element of the list would ever be evaluated)

aadler: First, that is why I am presenting this as a second option, one that is not deliberately crippled. This **is** the canonical implementation in R, where vectorization allows the loops to be peformed at the C/Fortran level and not the R level. The current implementation is similar to having the algorithm written in C, but requiring a script to select letters out of a pool to construct the call letter-by-letter each time—a deliberate crippling. If you want to properly compare languages, you need to write them in the language-appropriate method.

Regardless, I am not asking to remove or replace the current version, but to add this alongside to demonstrate R's speed when used **properly** and not in a way it was never meant to be used.

julian-urbano: Either way, what influences the time here is not so much the loop but the operations *inside* the loops. Right now they do the modulo operation `j %% u`, but we could do something simpler like addition `j %% u` and the results would be completely different. Compare e.g. Python and R:

|          | Python | R   |
|----------|--------|-----|
| Modulo   | 69     | 122 |
| Addition | 62     | 47  |

Yes, the modulo operation is notoriously slow in R...but that doesn't mean that loops are slow.

bddicken:
I am not an R expert, but as @axman6 pointed out, this seems like it's just fundamentally different. Is it much faster? Probably! But it also doing way less "work" than the other programs. We could similarly optimize all the other solutions to do less work like this, but that's not exactly the point of this.

I appreciate you submitting and providing this alternative implementation, but it isn't suitable for what this specific `loops` program is for.

J-Moravec:
@bddicken In all honestly, the current code is not representative of R because it is performing a cardinal sin of not vectorising see [R inferno](https://www.burns-stat.com/pages/Tutor/R_inferno.pdf).

R basic types are vectors, it doesn't have single-element scalar types like `int`, `double` or `char`. Because of this, operations on single elements are inefficient.

No one does:

```r
bad_sum = function(x){
sum = 0
i = 0
n = length(x)
  while(i < n){
    i = i + 1
   sum = sum + x[i]
  }
  sum
}
```

because it would be slow, inefficient, and honestly terrible code. Instead, people use `sum` that does that much faster (by doing the above loop within the compiled C code instead interpreted R code).

For this reason, the operations required for the above R loop are not even optimized, it is assumed that you will properly vectorise your function, use the in-build `sum`, `mean`, `var`, `rowsums` and `colsums` for matrices etc.

And even if you can't really use the optimized operations, the R standard way to do stuff is to still use various types of maps like `apply`, `lapply`, `sapply`, vapply`, that while being run in the interpreter and not the compiled lang, are the standard way to write operations.

aadler:
@bddicken I was asking to change the current deliberately retarded method you have but to add this alongside to generate a proper comparison. The way to compare languages is to set **tasks** and then allow each language to fulfill the task in its own best way. This means vectorized calculations for R, list comprehensions or generators for Python, etc. Forcing each language to perform a pre-set algorithm is simply foolish. Oh well, I guess someone will have to fork this repo and then allow each language to perform as was intended within that language. @axman6, any interest?

jpablo:
Maybe a good compromise would be to report two versions: the current "naive" algorithm (i.e. what someone not familiar with the language but with an imperative background would write) and an optimized version.

aadler:
That’s what I suggested and why my code has a different name than the existing one.

PEZ:
A way to look at the **loops** benchmark is that it shows how not to do things in R. 

I think a comment in the existing solution code could highlight that the implementation is not idiomatic R. The comment could also mention how it is solved in proper R code.

PR #177, opened 2024-12-04
Update code.R

LouisZAM:
avoid fetching element in vector each iterations.


aadler:
See #175 for a more R-appropriate implemenation.

bddicken:
As with #175, this is fundamentally doing a different thing. We *could* change all the other implementations to do this too, and some of those would see a nice speedup as well.

Thoughts?

bddicken:
Have not heard back from the author, so closing.

PR #178, opened 2024-12-04
Julia: update benchmark code

non-Jedi:
Please see each commit message for justification.

### DIFF
--- a/fibonacci/julia/code.jl
+++ b/fibonacci/julia/code.jl
@@ -1,15 +1,5 @@
-# base case 1
-fibonacci(::Val{0}) = 0
-# base case 2
-fibonacci(::Val{1}) = 1
-# general case
-fibonacci(::Val{n}) where n = fibonacci(Val(n-1)) + fibonacci(Val(n-2))
+fibonacci(n) = n < 2 ? n : fibonacci(n-1) + fibonacci(n-2)
 
-let
-    u = parse(Int,ARGS[1])
-    r = 0
-    for i ∈ 1:u
-        r += fibonacci(Val(i))
-    end
-    println(r)
-end
+main(u) = println(sum(fibonacci, 1:u))
+
+isinteractive() || main(parse(Int, ARGS[1]))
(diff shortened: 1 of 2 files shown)


fibonacci/julia/code.jl
```diff
@@ -1,15 +1,5 @@
-# base case 1
-fibonacci(::Val{0}) = 0
-# base case 2
-fibonacci(::Val{1}) = 1
-# general case
-fibonacci(::Val{n}) where n = fibonacci(Val(n-1)) + fibonacci(Val(n-2))
+fibonacci(n) = n < 2 ? n : fibonacci(n-1) + fibonacci(n-2)
```
bddicken, line 1: This needs to go back to having an if / case for 0 and 1.

fibonacci/julia/code.jl
```diff
@@ -1,15 +1,5 @@
...
-let
-    u = parse(Int,ARGS[1])
-    r = 0
-    for i ∈ 1:u
-        r += fibonacci(Val(i))
-    end
-    println(r)
-end
\ No newline at end of file
+main(u) = println(sum(fibonacci, 1:u))
```
bddicken, line 3: Can you explain if / how this is equivalent to the loop?

fibonacci/julia/code.jl
non-Jedi: Is `(n == 0 || no == 1) ? n : fibonacci(n-1) + fibonacci(n-2)` okay, or does it need to be the full

```julia
n == 0 ? 0 :
    n == 1 ? 1 :
    fibonacci(n-1) + fibonacci(n-2)
```

fibonacci/julia/code.jl
non-Jedi: The `sum` function just calls `mapreduce` with addition as the reducing function. So this is equivalent to mapping `fibonacci` to all elements in the range from 1 to `u` and then adding them up.

If you're prioritizing making the implementations more uniform between languages rather than using more idiomatic language constructs, going back to a loop is just fine. Let me know.

PEZ:

* #371
* #410

The new reference fibonacci programs do `if (n < 2) return n` for early termination, btw. Or whatever is faster for each language.

PR #180, opened 2024-12-04
Optimize Python recursive Fibonacci using LRU cache

Karatsidis:
This PR improves the recursive Fibonacci implementation by adding memoization using Python's built-in functools.lru_cache. 

This optimization reduces redundant computations and lowers time complexity from O(2^n) to O(n) while preserving the original recursive structure.

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -1,5 +1,7 @@
 import sys
+from functools import lru_cache
 
+@lru_cache(maxsize=None)
 def fibonacci(n):
   if (n == 0):
     return 0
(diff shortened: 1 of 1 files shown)


axman6:
As with the other solution you posted, this clearly changes the implementation from being `O(fibonacci(N))` to `O(N)`, and seems to miss the point of the benchmark - this just hands off the explicit array to a library that does the same thing.  From the README:

# fibonacci

Emphasizes function call overhead and recursion.

bddicken:
Seems like this will cause it to use cached results of the fibonacci computation? If so, this can't be accepted ass it fundamentaily (and significantly) changes how much work the program is doing. Thanks for submitting though!

PR #181, opened 2024-12-04
Go: change rand func to Int31n

gabstv:
Int31n returns, as an int32, a non-negative pseudo-random number in the half-open interval [0,n).

This change will avoid the  unnecessary casting (generating a low i64 integer number then casting to i32)


bddicken:
nice little fix!

PR #182, opened 2024-12-04
Add Chez Scheme

williewillus:
Add Chez Scheme

Chez Scheme (https://github.com/cisco/ChezScheme) is a Scheme that primarily implements
the R6RS standard of Scheme. It's well-known as one of the fastest Scheme implementations,
and also serves as the runtime implementation of other languages such as Racket CS
and Idris.

Note that on some machines, Chez Scheme is installed as `scheme` or `chezscheme`. On the
author's platform (Homebrew) and mine (Arch Linux), Chez Scheme is installed as `chez`, so
that's what I use.

Below are my machine's numbers.

Loops:
time chez --program code.so 40
1952387

real    0m11.646s
user    0m11.626s
sys     0m0.013s
(vs GCC 14.2.1 O3 1.948s real)

Fibonacci:
time chez --program chez/code.so 40
165580140

real    0m1.014s
user    0m0.986s
sys     0m0.027s
(vs GCC 14.2.1 O3 0.407s real)


williewillus:
Forgot to talk about the implementation strategy, but it's basically translated directly from the C code.

bddicken:
Cool! Thanks for this.

PR #183, opened 2024-12-05
Add F# and tweak existing C# submission

neon-sunset:
## Description

This introduces the following changes:
- Adds F# submission
- Renames C# projects to `code` to be consistent with other submissions
- Adds `OptimizationPreference=Speed` property to AOT mode for .NET projects, it is the `-O3` flag for NativeAOT

## Discussion

I have noticed that Dart places first on the diagram. Dart also has to work around its compiler being unable to infer side-effects by manually triggering them with preliminary `a[i];`. I believe this should not be allowed and the languages should be judged by their normal implementation. Coincidentally, it is also placed first in the run.sh script. If these benchmarks are ran on a thermally-constrained environment like MacBook Air which does not have active air cooling - it will give unfair advantage to the benchmarks of the languages that run first and is unlikely to be suitable for this scenario. The `--runs 3` value may also be too low and subject the submissions to more variance than desired. I understand that there are many languages and it takes long time to run, but given how close every one is placed on the `loops` benchmark - it increases in relevance.

Thanks!


bddicken:
Cool, thanks for the change and the discussion points!

The change with the `a[i];` came from here: https://mrale.ph/blog/2024/11/27/microbenchmarks-are-experiments.html

At first I was hesitant to include it, but decided it was acceptable, but of course this subject could be debated.

PEZ:
Please consider helping to port C# over to the new runner:

* https://pez.github.io/languages-visualizations

PR #194, opened 2024-12-05
Add implementation (for loops) using numba in python

mgzuber:
I'm not sure this is worthwhile or not, I'm happy to have this PR be rejected, but I did think it was worth it to show that python can be speed up with open source tools. Numba is great for the looping benchmark, and on my machine it speeds up the benchmark by 22x.

Numba does not currently support python 3.13, so it 3.12 needs to be used for now and numba needs to be installed.


mgzuber:
I see there is another PR with this already. I will close this one

PR #195, opened 2024-12-05
fix(Python fibonacci): adding cache to speed up the calculation

DavidVujic:
(also: code formatting)

Python is well-known for not being optimized for recursions like this. 

A "hack" is to use the builtin `functools.cache` that speeds up this kind of thing significantly. I _think_ this is the way the Pythonistas would write a function like this.

I understand that this is not the intention of the benchmark, but wanted to highlight that you wouldn't write Python in that kind of way because of the limitations of it 😄

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -1,17 +1,22 @@
 import sys
+from functools import cache
 
+
+@cache
 def fibonacci(n):
-  if (n == 0):
-    return 0
-  if (n == 1):
-    return 1
-  return fibonacci(n-1) + fibonacci(n-2)
+    if n == 0:
+        return 0
+    if n == 1:
+        return 1
+    return fibonacci(n - 1) + fibonacci(n - 2)
+
 
 def main():
     u = int(sys.argv[1])
     r = 0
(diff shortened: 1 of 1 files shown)


mgzuber:
All language implementations for the fibonacci benchmark are the naive version, so this isn't a like for like comparison to other languages

axman6:
> # fibonacci
>
The point of the benchmark isn't to change the `O(Fibonacci(N))` algorithm into an `O(N)` one, it's to do the work of the `C(Fibonacci(N))` algorithm. If this is something python isn't great at, then the benchmark should show that - most languages struggle with the naïve implementation, and all languages can implement the linear time version that this change changes it to, but that's not the point.

bddicken:
While I do appreciate this submission, I also think that this is an unreasonable change. By caching the function results, were allowing python do do *significantly* less computational work than the other languages.

DavidVujic:
I understand, no worries. What I wanted to show was that even though Python is not optimized for this kind of recursive calls, there's often other ways to solve problems like this one (by using the 🔋 🔋 included in the language).

PR #197, opened 2024-12-05
Kotlin has tail recursion

morisil:


PEZ:
I'm thinking that this is not what is meant by a **naïve fibonacci** implementation. 😄

fibonacci/kotlin/code.kt
```diff
@@ -1,7 +1,7 @@
-fun fibonacci(n: Int): Int {
-    if (n == 0) return 0
-    if (n == 1) return 1
-    return fibonacci(n-1) + fibonacci(n-2)
+tailrec fun fibonacci(
+    n: Int, a: Int = 0, b: Int = 1
+): Int = when (n) {
+    0 -> a; 1 -> b; else -> fibonacci(n - 1, b, a + b)
 }
```
He-Pin, line 5: Scala has too but we can't make use of it:(

axman6:
> # fibonacci
>

The benchmark is intended to test the naïve implementation in each language, not the fastest way to compute fibonacci numbers.

bddicken:
Thanks for submitting this.

However, we could make many of the other languages faster as well by using tail recursion. However, I want all of the languages to use the naive solution, which provides better stress testing of function calls and stack.

PR #198, opened 2024-12-05
⚡️Optimize the slice access in Golang

roma-glushko:
Optimizing the slice access in the golang loop source code per this subreddit conversation:

https://www.reddit.com/r/golang/comments/1h150ho/go_is_slower_than_java_on_same_code/

Golang's compiler doesn't do much optimization around slice access which leads to awful benchmark results.

### DIFF
--- a/loops/go/code.go
+++ b/loops/go/code.go
@@ -1,22 +1,32 @@
 package main
+
 import (
-    "fmt"
-    "math/rand"
-    "strconv"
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for submitting this @roma-glushko. As @karololszacki pointed out, we already have another PR with a similar change and some discussion, so I ill close this one out.

PR #202, opened 2024-12-05
Add V implementation

spytheman:
V (or Vlang) can be installed from its source, following these instructions:
```sh
git clone --depth=1 https://github.com/vlang/v
cd v
make
```

The performance of the resulting executables, on my machine (Intel i3) on Ubuntu 20.04, is ~ the same as C:
image


spytheman:
Here is the result for the second `loop` benchmark (it took a lot longer to run than the first one). V is again ~ the same as the C version (which is not surprising, since it generates virtually the same C code, for the user functions):
image

bddicken:
I will get to this either tomorrow or over the weekend. Have had other things going on, but thanks so much for submitting this.

bddicken:
Nice, thanks!

PR #203, opened 2024-12-05
Optimize Pascal implementations and add Delphi benchmarks

oldfort76:
- Added Delphi implementations for comparison:
  * Fibonacci: 871.8ms
  * Loops: 1.719s

- Added optimized Free Pascal implementations:
  * Fibonacci: ~15.6% faster (1.039s -> 876.7ms)
  * Loops: Similar performance with better type safety

Changes include:
- Changed to Cardinal type for unsigned integers
- Added compiler optimization directives
- Optimized parameter parsing
- Added project files with optimal compiler settings
- Added documentation with benchmarks and settings

Performance measured on Windows with both Delphi 2007 and FPC.


oldfort76:
Updated the compile script for free pascal (using the optimizations that are always used in an actual release: there's no point in sending the debug version for production). And updated the run script to run the delphi exe. I suppose it's hard to duplicate the results since the user will need to compile himself (as I cant upload a exe here)... But if someone has Delphi, even if a newer version, it will work.

PR #206, opened 2024-12-05
Interchange loops for efficiency in C "loops"

mmurrian:
These changes demonstrate a significant speed-up in the C loops benchmark while calculating otherwise identical results.

'a' can be initialized to r instead of 0. This avoids an unnecessary initialization and allows the 'i' and 'j' loops to be interchanged. And with the loops interchanged, the relatively expensive "j%u" operation can be more readily optimized by the compiler. In source code, there are still 1 billion "+= j%u" operations, same as the current benchmark.

I don't necessarily expect this pull request to be accepted because you might deliberately want the benchmark implementations to be "identical". But if you're open to a faster solution to the same problem, here it is!

Thanks!

### DIFF
--- a/loops/c/code.c
+++ b/loops/c/code.c
@@ -7,12 +7,14 @@ int main (int argc, char** argv) {
   int u = atoi(argv[1]);               // Get an input number from the command line
   srand(time(NULL));                   // FIX random seed
   int r = rand() % 10000;              // Get a random integer 0 <= r < 10k
-  int32_t a[10000] = {0};              // Array of 10k elements initialized to 0
-  for (int i = 0; i < 10000; i++) {    // 10k outer loop iterations
-    for (int j = 0; j < 100000; j++) { // 100k inner loop iterations, per outer loop iteration
-      a[i] = a[i] + j%u;               // Simple sum
+  int32_t a[10000];                    // Array of 10k elements initialized to a random value
+  for (int i = 0; i < 10000; i++) {
+    a[i] = r;
+  }
+  for (int j = 0; j < 100000; j++) {   // 100k outer loop iterations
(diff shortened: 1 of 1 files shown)


axman6:
Changing the work that's done by rearranging the loops seems to defeat the purpose of the benchmark - the point is not to get the correct result as quickly as possibly, it's to do the work as quickly as possible. You can calculate each `a[i]` in constant time knowing `r`, and those results could easily be computed at compile time, but that's not the point.

mmurrian:
Both u and r need to be known to calculate a[i]. r is a run-time random integer between 0 and 9,999. u is a user-provided 32-bit integer. So, I'll suggest a[i] is not actually so "easy" to compute at compile time. If it were so "easy", surely one of the optimizing compilers would have noticed and pre-computed a look-up table. None apparently did. Leaving your hyperbole aside...

If the point is to do "the work as quickly as possible", and that means something different than getting "the same observable result as quickly as possible", then I must inform you that optimizing compilers are defeating the purpose of the benchmark. They do precisely that: reduce the amount of work done to get the same observable result.

Given that, I can only assume that point of the benchmark is to compare how well various compilers manage to optimize a (deliberately?) poorly structured algorithm. That's unfortunate because that isn't how these benchmark results are being presented, or interpreted, on social media.

My code still performs largely the same amount of "work" at the language level: there are 1 billion iterations of the inner loop and 1 billion additions of j%u. The difference in initializing a[i] = r versus adding it at the end is negligible.

All that to say, I acknowledged above that I didn't expect this PR to be accepted. Rather, I thought it might be educational for less experienced devs to see a few very typical optimizations that provide significant performance improvements.

PR #208, opened 2024-12-05
Added COBOL loop impl

FinnDore:
I added the loop implementation for Cobol. It's not as fast as you would hope and I don't have a mainframe free to test so GNU Cobol has to suffice.

### deps
`brew install gnucobol gcc` or `apt install gnucobol libgmp-dev gcc`

```
Benchmarking COBOL
Benchmark 1: ./cobol/main
  Time (mean ± σ):     122.890 s ±  0.898 s    [User: 120.062 s, System: 1.142 s]
  Range (min … max):   122.066 s … 123.847 s    3 runs
```


bddicken:
Cool!

PR #209, opened 2024-12-05
Add loop in HolyC

orhun:
Hey! 👋🏼 

> HolyC is a programming language created by Terry A. Davis, the creator of the TempleOS operating system.[\*](https://en.wikipedia.org/wiki/TempleOS)

This PR adds the HolyC version of the loops program. I have no idea how to run or benchmark it outside of TempleOS but it was definitely fun!

templeos

templeos2

As you can see from the last line, it takes around 6 seconds in a single-core VM. I haven't run it on bare metal yet, and I'm not sure if I will ever do that.

Just wanted to share it here in case someone else is also interested in benchmarking HolyC.

This was done in a livestream, you can watch the recording [here](https://www.youtube.com/watch?v=QPLFHbZTEhk).

Cheers!

### DIFF
--- a/loops/holyc/code.HC
+++ b/loops/holyc/code.HC
@@ -1,0 +1,16 @@
+U0 Main(U32 n) {
+  U32 r = RandU32 % 10000;
+  U32 *a = MAlloc(sizeof(U32)) * 10000);
+  U32 i;
+  U32 j;
+  for (i = 0; i < 10000; ++i) {
+    for (j = 0; j < 100000; ++j) {
+      a[i] = a[i] + (j % n);
+    }
+    a[i] = a[i] + r;
+  }
+  "%d\n",a[r];
+  Free(a);
+}
+
+Main(10);


bddicken:
Cool. Can this run on a mac or ubuntu environment?

orhun:
Thanks! As I have said, I couldn't find a compiler that works fine on Arch Linux, but these re-implementations worth a try:

- https://github.com/nrootconauto/HolyCC2
- https://github.com/Jamesbarford/holyc-lang
- https://holyc-lang.com/

There isn't seem to be an online compiler :/

Or you can run the operating system via QEMU for the true experience, it's only 16MB!

Maybe some HolyC expert would know better about this though. I'd recommend waiting on this a bit :)

bddicken:
While this is cool, I don't want to include a language with a compiler that is hard to get working on various platforms. Closing it out.

orhun:
Understandable :) I had fun hacking this together either way!

PR #213, opened 2024-12-06
Update code.py to run faster

eduardopaula:
Changes Made:

- Elimination of Nested Loops:  
  Replaced the "for" loops with vectorized operations using NumPy, which are significantly faster.

- Use of NumPy for Mathematical Operations:  
  The modulus and summation operations were performed on a vector using NumPy ("np.arange" and ".sum()").

- Efficient Array Creation:  
  Used "np.full" to create the array initialized with the previously calculated sum value.

- Vectorized Randomization and Summation:  
  Added the random value "r" directly to the array "a" using vectorized operations.

### DIFF
--- a/loops/py/code.py
+++ b/loops/py/code.py
@@ -1,13 +1,25 @@
 import sys
 import random
+import numpy as np
 
 def main():
     u = int(sys.argv[1])         # Get an input number from the command line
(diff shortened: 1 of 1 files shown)


bddicken:
Thanks for submitting, however this significantly alters the amount of "work" the program is doing. We need to do ~ the same ~ amount of work in all programs for this to be reasonably fair.

PR #217, opened 2024-12-06
Significantly faster C benchmark, minimal change

mmurrian:
I've seen other pull requests that propose much larger changes than this to their respective benchmark. There is one for C implemented using AVX2 SIMD intrinsics. Another for Python restructures the loops and makes Numpy library calls, etc, to allow for vectorization.

I share this implementation because, as you can see, it makes no actual changes to the algorithm itself. By simply pre-computing "j%u" in advance, the benchmark can run **much** faster and SIMD auto-vectorization is now trivial for the compiler (no platform-specific SIMD intrinsics required).

32 times faster with AVX2 enabled (clang -mavx2)
70 times faster with AVX512 (clang -mavx512f)
10 times faster than the other PR using AVX2 SIMD intrinsics

As with my other PR that I closed, I don't necessarily expect this to be accepted because I understand the intent for "the same" implementations between languages. But I'd venture to say that this is the fastest C benchmark that's closest to the original. So, I share it.

Regards.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -1,4 +1,4 @@
-clang -O3 c/code.c -o c/code
+clang -march=native -O3 -Ofast -o c/code c/code.c
 clang++ -std=c++23 -march=native -O3 -Ofast -o cpp/code cpp/code.cpp
 go build -ldflags "-s -w" -o go/code go/code.go
 javac jvm/code.java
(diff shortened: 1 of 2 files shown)


bddicken:
Cool change, but this is significantly altering the amount of work being done here. We could make a similar change in all the other languages and give those a boost as well, but I'm not going to do that at this time.

PR #218, opened 2024-12-06
Publish C# and F# in release mode

agailloty:
Default publish mode is Debug, not all optimisations are made


neon-sunset:
Default publish mode is no longer `Debug` starting with .NET 8.

agailloty:
> Thankfully, the default publish mode had been switched to `Release` starting with .NET 8 😄

Oh thank you for mentioning that !! I didn't know

PR #219, opened 2024-12-06
Add Octave (MATLAB) implementation

Karatsidis:
- MATLAB remains widely used in academic research, engineering, and prototyping, particularly for mathematical modeling and algorithm development in universities and spin-offs. 
- Octave Compatibility: As Matlab is paid and proprietory, we choose the free / open-source alternative which is Octave. 
The implementation ensures accessibility for users without MATLAB licenses. Octave offers similar syntax and functionality for scientific computing tasks.
- Note that Octave is generally slower than Matlab; so results cannot be directly extrapolated to Matlab.


bddicken:
Nice, thanks for adding this.

PR #223, opened 2024-12-07
Babashka: Use separate code from Clojure

PEZ:
Babashka is the scripting version of Clojure. These changes take advantage of that to create simpler and more straight forward implementations, that are purely scripts.

Also: In Babashka there generally is no performance price to pay for using immutable data structures, even often the opposite. For the loops benchmark these changes give an almost 50% reduction in execution time.


fibonacci/bb/code.clj
```diff
@@ -0,0 +1,14 @@
+(defn- fibonacci [n]
+  (if (<= n 1)
```
bddicken, line 2: Can we make this two separate if conditons please?

PEZ: Done.

bddicken:
Nice! Merging.

PR #224, opened 2024-12-08
Add gleam benchmark

lemorage:
This PR adds benchmarks for gleam language. Since the `code` filename caused **Erlang module name collision**, so another filename was used.


PEZ:
Hello. Sorry for letting this hang for so long! I haven't seen it until now.

We have changed the way we run benchmarks:

* #371 

To get Gleam in, it should be implemented targeting the new runner. Let me know if you're up to it. I promise to be very responsive with feedback and eager to merge.

PR #230, opened 2024-12-10
added fibonacci cpp code

DuarteCruz6:
created the fibonacci c++ code. 

i have a question tho, should i use cout or printf?
printf is way faster, but it is from C tho. cout is the main C++ one


bddicken:
Thanks!

Issue #220, opened 2024-12-06
The benchmark penalizes languages that need to start a runtime environment (e.g. anything on the JVM or .Net)


PEZ:
Hi,

Thanks for creating this project, @bddicken! 🙏 

I have been running benchmarks in some different ways the last few days. I notice that a lot of the difference between the Java performance and C is about JVM start times. And for Clojure the penalty is much bigger, because then also Clojure needs to start before the benchmarks can start crunching. To “see” the start time I created a `hello-world` benchmark, where the programs only print “Hello World!”. Here's a run on my Mac M4:

```
Benchmarking C
Benchmark 1: ./c/code 40
  Time (mean ± σ):       2.2 ms ±   0.2 ms    [User: 0.7 ms, System: 1.2 ms]
  Range (min … max):     2.0 ms …   2.5 ms    10 runs
 

Benchmarking Java
Benchmark 1: java jvm.code 40
  Time (mean ± σ):      38.7 ms ±   0.9 ms    [User: 14.1 ms, System: 21.4 ms]
  Range (min … max):    37.3 ms …  40.2 ms    10 runs
 

Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.12.0/clojure-1.12.0.jar:/Users/pez/.m2/repository/org/clojure/core.specs.alpha/0.4.74/core.specs.alpha-0.4.74.jar:/Users/pez/.m2/repository/org/clojure/spec.alpha/0.5.238/spec.alpha-0.5.238.jar code 40
  Time (mean ± σ):     275.0 ms ±   4.3 ms    [User: 410.4 ms, System: 43.4 ms]
  Range (min … max):   270.0 ms … 285.0 ms    10 runs
 

Benchmarking Babashka
Benchmark 1: bb -cp clojure -m code 40
  Time (mean ± σ):      11.9 ms ±   0.4 ms    [User: 3.3 ms, System: 6.1 ms]
  Range (min … max):    11.1 ms …  12.4 ms    10 runs
```

Basically the JVM takes 40ms to start, and then Clojure takes another 240ms. To be compared with the benchmark times for e.g. C, Java, and Clojure, which run at 250, 350, and 550 ms, respectively on the same machine.

[...]

Issue #157, opened 2024-12-03
Replace entire suite with one that has already been golfed to death


headius:
I do not believe this suite provides anything better than the many existing benchmark suites out there. There's currently two benchmarks: loops and fibonacci. Neither provides any useful performance metric for any tested language. The benchmark harness itself just runs each implementation once from a command line, which prevents many optimizing runtimes from optimizing any code, and includes startup time for all runtimes.

There are many other benchmark suites that have already iterated over similar benchmarks (plus include several that do real work), such as the Benchmarks Game: https://benchmarksgame-team.pages.debian.net/benchmarksgame/index.html

Individual language implementations also frequently have more complete suites of representative microbenchmarks, such as for [PyPy](https://speed.pypy.org) and Ruby's [YJIT](https://speed.yjit.org)

And there are many existing harnesses for benchmarks that can deliver more meaningful results. They do things like run multiple times, run repeatedly in a single process, and isolate the execution of the runtime and the benchmark harness from timing results.

By having yet another microbenchmark suite that doesn't measure anything useful and does it in the most inaccurate way, this repository is creating work for language and runtime implementers that now have to answer questions about "why aren't you fast" and submit PRs to fix benchmarks that are useless to begin with.

I would recommend replacing this suite with one that has valuable benchmarks run in a meaningful way, or else just delete it.

briansterle:
there's also Dave Plummer's [prime sieve benchmark](https://github.com/PlummersSoftwareLLC/Primes)

bddicken:
Thanks for the feedback @headius. I'm sure you know far more about benchmarking and compilers than I do, so it means a lot to get your input like this. I'll reply to a few of the things you brought up:

> I do not believe this suite provides anything better than the many existing benchmark suites out there.

To be honest, this kinda snowballed (as you might have seen on twitter). I started by getting just a modest number of PRs, and then after I published are round of results I got over 100 and then things just kinda kept rolling. I actually think it's pretty cool to be able to crowdsource a language comparison like this, even if it isn't perfectly academic, which it certainly is not (I worked in academia for 7 years!).

> The benchmark harness itself just runs each implementation once from a command line,

We actually do warmup runs and multiple timed executions, using hyperfine. If you have Ideas for how to configure hyperfine differently, please make a PR.

> and includes startup time for all runtimes.

[...]

PR #231, opened 2024-12-11
Add hello-world benchmark

PEZ:
Hello,

I started to add a hello-world benchmark. Two drivers for me:

1. To have a benchmark that shows how quick a program written in a language is to start. Related to:
   * #220 
1. To learn how Hello World looks in all these cool languages. And what the toolchains look like.
   * Since I picked up Clojure, I have stopped testing other languages almost completely, so it is probably healthy as an exercise.

When it comes to toolchains I at some point decided I wanted CI to run the benchmark. Making all these toolchains co-exist in a Github Actions workflow is VERY time consuming, though, so it put a bit of a stop in my implementing of Hello Worlds. But it is probably good to have one fast-running benchmark that can test the clean-compile-run of all languages reliably.

Anyway, @bddicken, before I prioritize this PR too much, I'd like your feedback on the idea and the way I am going at it. Is it something you would even consider merging? I'll have use for it even if you don't so don't hesitate to tell me the truth. 😄 

As for the state of this PR, I have added some 15 languages so far, maybe less, maybe more. And CI runs some of them. I think my next steps will be to add more languages and then see how many of them I can get to run in CI.


bddicken:
This is actually a very cool idea. One of the big complains outlined in [this issue](https://github.com/bddicken/languages/issues/157) was that this test suite includes startup times, so it would be super cool just to see what the cost actually is for all of them!

bddicken:
Is this ready to merge, or do you want to work on it more first?

PEZ:
Good question! I was thinking to just show you what I was up to, but I do also think it can be merged and we can take it from there. Maybe other people will want to add languages and help with the CI setup. 

The thing to keep in mind is that it's not all put together in CI. So there are languages that have hello-world, that won't have a tool chain set up in CI. But I guess that's fine as long as you are aware.

bddicken:
Ok great. I'll just go ahead and merge this as a start, and others will be welcome to add!

PEZ:
> One of the big complains outlined in [this issue](https://github.com/bddicken/languages/issues/157) was that this test suite includes startup times, so it would be super cool just to see what the cost actually is for all of them!

Indeed. Seeing the startup times was my main driver for this. Let's see if it gives us any ideas how to mitigate. Hello World is not dealing with the JIT aspects, though. But maybe we can figure that out too, eventually.

igouy:
> I started to add a hello-world benchmark. 

I had kind-of hoped people wouldn't just repeat the same mistakes:

https://salsa.debian.org/benchmarksgame-team/archive-alioth-benchmarksgame/-/tree/master/contributed-source-code/shootout/hello

igouy:
> One of the big complains outlined in [this issue](https://github.com/bddicken/languages/issues/157) was that this test suite includes startup times, so it would be super cool just to see what the cost actually is for all of them!

So measure it ?

https://benchmarksgame-team.pages.debian.net/benchmarksgame/performance/startup.html

PR #233, opened 2024-12-11
Add php levenshtein.

KsaR99:


levenshtein/php/code.php
```diff
@@ -0,0 +1,31 @@
...
+// Main function to mimic C program's behavior
+function main(array $args): void
+{
+    $minDistance = -1;
+    $times = 0;
+
+    foreach ($args as $i => $str1) {
+        foreach ($args as $j => $str2) {
+            if ($i !== $j) {
+                $distance = levenshtein($str1, $str2);
```
bddicken, line 14: We need to actually implement `levenshtein` in this code rather than relying on the standard library function. It's possible they are not using the same exact algorithm as described in the [levenshtein readme](https://github.com/bddicken/languages/tree/main/levenshtein#readme). Can add an implementation here?

bddicken:
Thanks for submitting this! I left a note on the levenshtein program.

Also, can you please revert the changes to the other two programs and just use this for adding levenshtein?

levenshtein/php/code.php
KsaR99: @bddicken Ad.1 according to php "c"-source they behave the same (php have some improvements on memory management, but not a speed). For reference - https://github.com/php/php-src/blob/master/ext/standard/levenshtein.c Ad.2 Sure, I will refactor to pure implementation if still necessary ;

KsaR99:
I will do. BTW. Can these be made in separate PR then? for eg. commit #113e78766eabbe332ed99af1d73a0def4751f09f made "typehint's" that are not used without `declare(strict_types=1)`

levenshtein/php/code.php
bddicken: Interesting. I'd like to actually have the implementations in the repo so that we are actually seeing what the php interpreter / jit is capable of, rather than a pre-compile / optimized implementation in C

bddicken:
Yeah, you're welcome to make a separate PR for other changes. Though, some of it won't make it in, like the optimization to the fib function.

levenshtein/php/code.php
```diff
@@ -0,0 +1,62 @@
...
+
+declare(strict_types=1);
+
+function levenshtein_distance(string $str1, string $str2): int
+{
+    $m = strlen($str1);
+    $n = strlen($str2);
+  
+    // Create a matrix to store distances
+    $matrix = [];
```
bddicken, line 11: I didn't actually run this and I don't know php very well, but isn't this just a 1d array?

KsaR99: Because it's dynamic language. It is allocated to 2d right in below loops. Theres no way to

bddicken: 👍

PR #235, opened 2024-12-12
Add cpp levenshtein

studentenherz:
This is almost the same as the C example, just some cpp flavor on top but can be further improved.


bddicken:
Nice!

PR #237, opened 2024-12-12
Fix:  C# levenshtein.

AhmedZero:
in Top-level statements, C# index should start with 0 because the args don't include app path.
i'm sorry, that's my fault.


bddicken:
👍

PR #238, opened 2024-12-12
Add hello-world\php

KsaR99:
It using php "[short echo tags](https://www.php.net/manual/en/language.basic-syntax.phptags.php)", that are available since PHP 5.4 (01-march-2012).


bddicken:
NICE

PR #239, opened 2024-12-12
levenshtein\js; Use builit-in min() function

KsaR99:
Supported in deno, node etc. [doc reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min).


bddicken:
Very nice, TY @KsaR99.

PR #241, opened 2024-12-12
hello-world: add Ada

iharuspex:


PEZ:
Cool! Did you look at adding Ada to the CI toolchain? Otherwise I can have a look at that.

iharuspex:
Hi, @PEZ !
As far as I know, there is currently only one relevant way to configure GitHub Action for Ada - this is using the configuration from the [setup-alire](https://github.com/alire-project/setup-alire) repository. This configuration assumes the use of Alire for toolchain setup and build sources.
I can try to configure CI for Ada through this configuration, but this will require adding the file `alire.toml` to the directory with the `code.adb` to be able to run `gnatmake` through `alr exec`
Otherwise, I can try to fork some other repo that contains an older config that deploys only `gnat-native` toolchain and update it, but that's a more complex task that will take some time.

PEZ:
@iharuspex I note that currently the Ada compile is disabled in `compile.sh`:

```sh
#gnatmake -O3 -gnat2022 -gnatp -flto ada/code.adb -D ada -o ada/code
```

So probably @bddicken had some troubles with making it work?

If you can get setup-alire to work then we have a way to compile, which I think would be awesome.

I'm thinking that since Ada isn't compiled currently, adding the hello-world code alone is an improvement, and you can add the toolchain in a separate PR. That way we don't risk block this one by what could prove to be a bit of process to make work. (Speaking from the experience of adding some 20 toolchains. 😄)

iharuspex:
@PEZ 
I don't know how to install the GNAT Toolchain on MacOS without Alire. Linux systems usually have this toolchain in their repos. Anyway, I think that using Alire to install the toolchain is the best way, because it is a platform-independent way. :smile: 

Ok, so I'll experiment with this, and if everything works out, I'll create another PR with the addition of `alire.toml` to each of the test programs on Ada and with an updated CI configuration.

PEZ:
Sounds good. When I added the CI config, I started with ubuntu, but ran into problems (I have forgotten what, but generally more toolchains have homebrew recipes than have deb packages, for some strange reason). I got things to work using `macos` runner and figured that @bddicken also uses Mac so stayed with that. Ultimately I guess using a matrix of runners and have it work on all would be nice, but it already takes 6 minutes to run with only like 20% (maybe less) of the toolchains in place.  The runners will run in parallell, but my experience is that Windows will run 3X slower in CI. That would still leave it open for trying to get it to work in a matrix with Mac and Linux. Just jotting these thoughts down. It's for the long run I guess.

Let me know if you need any help testing things on Mac when you get there.

PEZ:
Let's worry about the toolchain in a separate PR.

iharuspex:
> Let's worry about the toolchain in a separate PR.

Ok, I'm working on it right now in a separate branch :)

PR #243, opened 2024-12-12
levenshtein: Make Babashka version clearer and more readable

PEZ:
- Using `as->` to make the nesting more shallow


bddicken:
Nice!

PR #222, opened 2024-12-07
Add cython implementation

poudro:
Thanks for the fun repo :)

This is pretty much cheating as this compiles down to C equivalent, but fun nonetheless

```
languages/loops/cython$ hyperfine -i --shell=none --runs 3 --warmup 2 "./code 40"
Benchmark 1: ./code 40
  Time (mean ± σ):     649.8 ms ±   0.2 ms    [User: 642.9 ms, System: 5.9 ms]
  Range (min … max):   649.6 ms … 650.0 ms    3 runs
```

As a short explanation, [Cython](https://cython.org/) is a python library that allows to write python code, or very similar, and convert it to C for efficiency.

PR #245, opened 2024-12-13
update python code

miraculixx:
I'm submitting this for a Python implementation that's both on par with C, using the original code in terms of code structure and logic (in particular, a nested for-loop), and is using only the Python standard library.

This is particularly addressing the concern raised in #213, #217 where you mention that the amount of (computation) work should be the same regardless of the language, and hence pre-computing `u % j` is not valid. Similarly, this is not using an external library (numpy), as you argued in #63 that only built-in / standard library features are allowed. 

**Approach**

- `code.py` is native python code, using the standard library; leveraging compute-parallelism and memoization
- `cythonized.py` is compiled to c using Cython, a python compiler

**Results**

- `code.py` runs at approx. 10x C speed (with memoization) to at 1/10x C speed (compute-parallel, without memoization). That's a 4x to 40x speed-up v.v. the original code.
- `cythonized.py` runs at roughly C speed, that's a 30x speed-up v.v. the original code.

**Rationale**

- memoization and compute-parallelism are commonly used built-in features of the Python standard library
- [Cython](https://cython.org/) is *an optimised static compiler for the Python programming language*

One could argue that memoization reduces the amount of work done; however, this is up to the program itself, i.e. the work *specified* by the program is the same regardless (unlike code-specified optimization, such as pre-computation or explicit caching). In some sense, this is semantically equivalent to other languages' compiler optimizations and use of just in time compilation (JIT).

The use of Cython to compile the code to native code is the same approach used by other languages. The program used for this (`cythonized.py`) is otherwise identical with the original code.


fibonacci/py/code.py
```diff
@@ -1,6 +1,7 @@
 import sys
+from functools import cache
 
-
+@cache
```
bddicken, line 4: Using a cache is "cheating" in this case. It allows the program to skip huge amounts of the work relative to other implementations. I mean, it's cool that python supports this, but a similar cache could be implemented in other languages giving them huge boosts also.

loops/py/code.py
```diff
@@ -1,16 +1,28 @@
...
+    r = random.randint(0, 10000) # Get a random number 0 <= r < 10k
+    a = [0] * 10000              # Create an array of 10k elements
+    # 10k outer loop iterations, in parallel and memoized
+    tasks = ((a[i], u, r) for i in range(10000))
+    a = list(parallel(compute, tasks))
+    print(a[r])                  # Print out a single element from the arra
 
+def parallel(fn, iterable):
+    # helper for parallel processing as a transparent fallback in case memoization is not effective
+    from multiprocessing import Pool
```
bddicken, line 23: multi threading and multi processing are not allowed for this particular benchmark. Though, if you have an idea for a multi-process-specific benchmark, feel free to make a PR to add a new one!

loops/py/cythonized.py
```diff
@@ -0,0 +1,24 @@
...
+@cython.locals(u=cython.int, r=cython.int, a=cython.int[10000], j=cython.int, i=cython.int)
+def main():
+    u = int(sys.argv[1])         # Get an input number from the command line
+    r = random.randint(0, 10000) # Get a random number 0 <= r < 10k
+    a  = [0] * 10000             # Array of 10k elements initialized to 0
+    for i in range(10000):       # 10k outer loop iterations
+      for j in range(100000):    # 100k inner loop iterations, per outer loop iteration
+        a[i] += j % u            # Simple sum
+      a[i] += r                  # Add a random value to each element in array
+    print(a[r])                  # Print out a single element from the arra
```
bddicken, line 19: I think I already merged a PR for cython?

fibonacci/py/code.py
miraculixx: Ok, wouldn't you have to disable all JIT and compiler optimizations too, because they are essentially "cheating" too? Otherwise the benchmark is not comparing *languages* but compilers.

loops/py/cythonized.py
miraculixx: There is PR #222, however that is treating Cython as a separate language (which is fair because it is using different syntax). This PR however uses Cython in its pure python mode.

poudro: > I think I already merged a PR for cython?

Actually, PR https://github.com/bddicken/languages/pull/222 hasn't been merged yet :)

loops/py/code.py
miraculixx: I thougt the objective is to benchmark 1 billion operations? I get that you want some ground rules, but why does it matter *how* that is achieved?

PEZ:
Closing this for now. As stated above already, we're not allowing memoization, caching, parallellism, or anything like that in the current benchmarks. I'd be interested in a cython entry. There's a new runner and any contributions need to target that.

PR #246, opened 2024-12-13
Update zig

Kayden1412:
- Use biased random number generator.
- Change print to writeAll.


bddicken:
I don't know zig very well. Can you explain why these changes are an improvement?

PR #247, opened 2024-12-13
fix(go): add a temp var to prevent reindex on each loop

omurilo:
A golang compiler don't optimize the runtime indexing of array `a`, so add a temporary variable to store the value of `a[i]`, compute values into the temporary variable and so on define new value for `a[i] = temp`


bddicken:
Thanks for contributing, but unfortunately this will not be accepted. This allows the program to do significantly less "work" thank the others. The same optimization *could* be made in the others to yield a performance improvement as well, but I will keep as-is for now.

If you have ideas for new benchmarks, please ket me know!

PR #250, opened 2024-12-14
Major improvements

Gkodkod:


Gkodkod:
I've made several optimizations to improve both time and space complexity:

**Space Optimization:**

- Reduced space complexity from O(m*n) to O(min(m,n)) by using only two rows
- Always use the shorter string for column dimension to minimize space usage
- Reuse arrays instead of creating new ones

**Time Optimization:**

- Optimized the main loop to avoid redundant comparisons (comparing string pairs only once)
- Removed unnecessary variable assignments and simplified calculations
- Used more efficient array access patterns

**Algorithm Improvements:**

- Implemented a more efficient version of the Wagner-Fischer algorithm
- Combined operations to reduce the number of array accesses
- Used direct Math.min calls instead of nested conditionals
- The space complexity is now O(min(m,n)) instead of O(mn), where m and n are the lengths of the input strings. The time complexity remains O(mn) for each comparison, but we've reduced the constant factors and improved cache locality.

For the string comparisons in main(), we now use a more efficient loop structure that avoids comparing the same pair of strings twice (i.e., we don't need to compare both (i,j) and (j,i) since the Levenshtein distance is symmetric).

These optimizations should provide better performance, especially for longer strings and larger sets of comparisons.

Gkodkod:
Prof. [Benjamin Dicken](https://benjdd.com/) @bddicken 

Instead of this PR See both of these with comprehensive comments :

https://github.com/bddicken/languages/pull/251
https://github.com/bddicken/languages/pull/253

Awesome job btw !!!  Mr. Gal Levinshtein @Gkodkod

PR #252, opened 2024-12-14
Make runner check for explicit executable

PEZ:
The runner doesn't currently work for all languages, because how they differ in what executable file they use. It's the thing we discussed here

* https://github.com/bddicken/languages/pull/234#issuecomment-2537409266

(Even if I don't think we agreed that things need to be fixed. 😄 Here's a fix anyway.)

This PR adds and argument to the run commands so that they point at the correct benchmark executable file (or script, when that's the case). Plus reverts the way that we currently split up the benchmark command and it's arguments. The run function now also reports when it can't find a compiled executable or script for the benchmark.

Please note that for some of the languages the actual compiled executable is mine and ChatGPTs best guess, so we probably need to verify this somehow. I believe you have almost all toolchains setup on your machine, @bddicken? If so maybe you are best positioned to verify. I am in the process of setting up more (hopefully all) toolchains, but it will take me a while.

Some more changes of the run.sh script:

* The input.txt file is now used for all benchmark. (Even if it is an empty file for hello-world.)
* The run.sh script takes arguments that can override the the input file. To make it easier to test the runner with faster executing arguments.


PEZ:
I've now merged with latest main. And also sorted the run.sh languages alphabetically, because it was quite hard to merge things and not miss something when things where unsorted.

bddicken:
I've made some changes myself to the run.sh script which addresses some of these issues. For example, I think the issue of knowing where the file to execute is is now resolved. Maybe if we do a zoom call we can chat about this then!

PR #253, opened 2024-12-14
Levenstein - 24 new languages implementations added

Gkodkod:
@bddicken @PEZ 

New directories created with improved code implementation for the Levenstein Lanaguages performance comparisons:

- awk
- chez
- common-lisp
- crystal
- d
- elixir
- fortran
- fsharp
- haskell
- julia
- lean4
- lua
- nim
- objc
- Ocaml
- odin
- perl
- r
- ruby
- rust
- scala
- swift
- typescript
- zig

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 *.class
 code
 .idea/
@@ -6,3 +7,12 @@ code
 *.ali
 .aider*
 .env
+node_modules/
+target/
+build/
+out/
+dist/
+*.log
+*.aux
+*.bbl
+*.blg
--- a/levenshtein/awk/code.awk
+++ b/levenshtein/awk/code.awk
@@ -1,0 +1,80 @@
+# Calculates the Levenshtein distance between two strings using Wagner-Fischer algorithm
+# Space Complexity: O(min(m,n)) - only uses two arrays instead of full matrix
+# Time Complexity: O(m*n) where m and n are the lengths of the input strings
+function levenshtein_distance(s1, s2,    m, n, prev_row, curr_row, i, j, cost, deletion, insertion, substitution) {
+    # Early termination checks
+    if (s1 == s2) return 0
+    if (length(s1) == 0) return length(s2)
+    if (length(s2) == 0) return length(s1)
+
+    # Make s1 the shorter string for space optimization
+    if (length(s1) > length(s2)) {
+        temp = s1
+        s1 = s2
+        s2 = temp
+    }
+
+    m = length(s1)
+    n = length(s2)
+
+    # Initialize first row
+    for (i = 0; i <= m; i++) {
+        prev_row[i] = i
+    }
+
+    # Main computation loop
+    for (j = 1; j <= n; j++) {
+        curr_row[0] = j
+
+        for (i = 1; i <= m; i++) {
+            # Get characters at current position
+            # Note: substr in awk is 1-based
+            cost = (substr(s1, i, 1) == substr(s2, j, 1)) ? 0 : 1
+            
+            # Calculate minimum of three operations
+            deletion = prev_row[i] + 1
+            insertion = curr_row[i-1] + 1
+            substitution = prev_row[i-1] + cost
+            
+            curr_row[i] = deletion
+            if (insertion < curr_row[i]) curr_row[i] = insertion
+            if (substitution < curr_row[i]) curr_row[i] = substitution
+        }
+
+        # Copy current row to previous row
+        for (i = 0; i <= m; i++) {
+            prev_row[i] = curr_row[i]
+        }
+    }
+
+    return prev_row[m]
+}
+
+# Main program
+BEGIN {
+    if (ARGC < 3) {
+        print "Please provide at least two strings as arguments."
+        exit 1
+    }
+
+    min_distance = -1
+    times = 0
+
+    # Compare all pairs of strings
+    # Note: ARGC includes the script name, so we start from 1
+    for (i = 1; i < ARGC; i++) {
+        for (j = 1; j < ARGC; j++) {
+            if (i != j) {
+                distance = levenshtein_distance(ARGV[i], ARGV[j])
+                if (min_distance == -1 || distance < min_distance) {
+                    min_distance = distance
+                }
+                times++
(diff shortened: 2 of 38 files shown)


He-Pin:
There is an allocation in Scala's impl, which makes it slow :(

https://contributors.scala-lang.org/t/remove-local-tuple-allocation/6969

Gkodkod:
@He-Pin Thanks for pointing this out and providing the informative link.  Great job on this !
https://github.com/bddicken/languages/pull/271

He-Pin:
@Gkodkod The Scala-js and Scala-native did more optimization in the backend, the code is right, but the Scala 3 compiler currently lacks this optimization:(

Gkodkod:
@He-Pin Take a look at these optimization ideas and the Scala 3 compiler enhancements. 

In the absence of compiler-level optimizations, as developers, we can manually refactor our code to avoid unnecessary tuple allocations. For instance, instead of creating a tuple for variable swapping, direct assignments can be used to achieve the same logic without the overhead.

Example: Instead of writing something like this...

`val (str1, str2) = if (s1.length > s2.length) (s2, s1) else (s1, s2)`

We can refactor to:

```
val isS1Bigger = s1.length > s2.length
val str1 = if (isS1Bigger) s2 else s1
val str2 = if (isS1Bigger) s1 else s2
```

This refactoring avoids the creation of an intermediate tuple, thereby reducing unnecessary allocations and improving performance.

Introducing a new phase in the Scala 3 compiler to specialize tuples, similar to how functions are specialized, could mitigate this issue. This approach is discussed in the Scala community as a way to enhance performance by reducing unnecessary allocations.

Take a look at this closed, and more detailed deep dive Scala 3's New Compile-Time Features video. 
https://github.com/scala/scala3/issues/11114
https://www.youtube.com/watch?time_continue=33&v=5PQYC0UDqGA&source_ve_path=MzY4NDIsMjg2NjY

Let us all know how we can improve it further, and thanks again.
 
Additionally, I have not gotten a chance to look at all the new optimizations and PR's, which I'm so delighted are taking place... Awesome progress @bddicken and @PEZ, and certainly a further indication how important this brillant idea is!

mriehl:
nice fortran "optimization", truly a Jia Tan moment

PEZ:
@mriehl, I am super curious about why Fortran so totally owns the competition in this benchmark, but I don't think it is due to some malicious intent from somebody. Can you elaborate on what you think is off with the Fortran contribution? To me it looks like a super straight-forward implementation of the reference implementation in C.

mriehl:
@PEZ yeah, I don't mean due to the malicious part, but once you see it, it really is sitting in plain sight.

I can't claim any credit for finding it, it was [investigated and shown by Casey Muratori in this stream](https://www.youtube.com/watch?v=uVVhwALd0o4).

I do recommend watching the entire second half as it has a lot of interesting considerations regarding this kind of benchmarking and exposes critical flaws in the methodology, but if you really just want to see the most immediate reason for the fortran performance it's at `?t=5376`.

PEZ:
I have a fixed version of the Fortran contribution on my machine now.

```patch
diff --git a/levenshtein/fortran/code.f90 b/levenshtein/fortran/code.f90
index 1c096cc..7e3a89e 100644
--- a/levenshtein/fortran/code.f90
+++ b/levenshtein/fortran/code.f90
@@ -90,8 +90,8 @@ program main
     implicit none
     
     integer :: num_args, i, j, distance, min_distance, times
-    character(len=100), allocatable :: args(:)
-    character(len=100) :: arg
+    integer :: max_len, arg_len
+    character(len=:), allocatable :: args(:)
     
     ! Get command line arguments
     num_args = command_argument_count()
@@ -101,13 +101,19 @@ program main
         stop 1
     end if
     
-    ! Allocate array for arguments
-    allocate(args(num_args))
+    ! Get max length of arguments
+    max_len = 0
+    do i = 1, num_args
+        call get_command_argument(i, length=arg_len)
+        max_len = max(max_len, arg_len)
+    end do
+
+    ! Single allocation of args array
+    allocate(character(len=max_len) :: args(num_args))
     
-    ! Read command line arguments
+    ! Read command line arguments directly into allocated array
     do i = 1, num_args
-        call get_command_argument(i, arg)
-        args(i) = trim(arg)
+        call get_command_argument(i, args(i))
     end do
     
     min_distance = -1
```

Gone is that crazy edge the Fortran contribution had over that of C! On my machine this makes Fortran end up second, behind Pascal.

```
Checking C
Check passed
Benchmarking C
  Time (mean ± σ):     374.4 ms ±   1.8 ms    [User: 366.6 ms, System: 3.7 ms]
  Range (min … max):   372.1 ms … 377.0 ms    7 runs
 

Checking Fortran
Check passed
Benchmarking Fortran
Benchmark 1: ./fortran/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     307.3 ms ±   2.8 ms    [User: 299.7 ms, System: 3.9 ms]
  Range (min … max):   304.0 ms … 312.7 ms    7 runs
 

Checking Free Pascal
Check passed
Benchmarking Free Pascal
Benchmark 1: ./fpc/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     206.0 ms ±   2.5 ms    [User: 200.1 ms, System: 3.1 ms]
  Range (min … max):   202.5 ms … 209.7 ms    7 runs
```

(This is run using my much larger input data.)

Will make a new benchmark run for https://pez.github.io/languages-visualizations/ tonight if I find the time.

cyrusmsk:
> this makes Fortran end up second, behind Pascal.
Did you fix Pascal bug before comparison? xD

PEZ:
> Did you fix Pascal bug before comparison? xD

No, that was before the Pascal bug was identified. I don't think it is fixed yet, btw.

> > this makes Fortran end up second, behind Pascal.

Also, that was compiling C with clang on my Mac. Compiling C with gcc makes it a tie between C and Fortran. Semi recent results here: https://pez.github.io/languages-visualizations/#levenshtein

PR #254, opened 2024-12-14
Clojure: Levenshtein updates

PEZ:
~I noticed that some Levenshtein implementations skip redundant comparisons in a way that the reference implementation does not.~ And also that some implementations calculates the wrong `min_distance`.

I think we can make the run.sh script check for correctness of the output. I can have a stab at that, if you like, @bddicken. The tricky one is `loops` with it's `random()`, but there's still a range we can check there I think.

UPDATE: Since most of my changes where about enforcing the old reference implementation, I have removed those now. And instead updated the Clojure implementation to match the new reference implementation.


bddicken:
Sweet!

PR #260, opened 2024-12-18
Disable "hello-world" workflow for (other) directories.

KsaR99:
Currently It runs on any commit to any file.

### DIFF
--- a/.github/workflows/hello-world.yml
+++ b/.github/workflows/hello-world.yml
@@ -4,9 +4,13 @@ on:
   push:
     branches:
       - '**'
+    paths:
+      - 'hello-world/**'
   pull_request:
     branches:
       - main
+    paths:
+      - 'hello-world/**'
 
 jobs:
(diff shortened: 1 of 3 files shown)


PEZ:
It should run for changes to `compile.sh` and `run.sh`. Maybe better with a blacklist of directories. So not run for files in `loops`, `levenshtein`, or `fibonacci`.

KsaR99:
- I added compile.sh & run.sh to white list.
- Also removed "components: 'native-image'" due to [issue in actions](https://github.com/bddicken/languages/actions/runs/12403096425)

issue



> Please remove "components: 'native-image'" from your workflow file. It is automatically included since GraalVM for JDK 17: https://github.com/oracle/graal/pull/5995

PEZ:
To me it makes total sense to do these changes, @bddicken. I don't think we'll be running the other benchmarks in CI anyway. At least not on all commits.

PR #262, opened 2024-12-18
Dart Refactor Levenshtein distance calculation in code.dart

Shreemanarjun:


PEZ:
Hi! What is the purpose behind these refactorings?

Ichoran:
The whole Dart implementation is wrong: it uses a full matrix instead of the two-alternating-rows algorithm in the reference implementation.  This refactoring does not help.

PR #265, opened 2024-12-19
Dart improvements to Levenshtein challenge, roughly 47% faster

vincevargadev:
This PR includes changes to the Levenshtein challenge, Dart solution.

The individual commits describe the steps:

* 6c9aeb6 Fix Dart code with standard dart formatting tool
* fc18e35 Use Int16List for inner int list
* 291bb16 Make outer list not growable
* 8f2059e Extract row and prev row in levenshtein dist calculation

The majority of the improvements are due to the inner list using [`Int16List`](https://api.dart.dev/stable/2.15.1/dart-typed_data/Int16List-class.html), though the other two commits (outer list not growable, and extracting the row into its own variable) also consistently performed a couple of % better.

I also formatted the code. In Dart, it's standard to use the [built-in Dart tooling for formatting](https://dart.dev/tools/dart-format) Dart source code.

On my machine, these changes make the Dart code roughly the same speed as Swift.

Languages that were previously faster, but are now (with my changes) slower than Dart: Java, LuaJIT, C#, Bun (using [this tweet](https://x.com/BenjDicken/status/1869412072318283783) as a reference).

### DIFF
--- a/levenshtein/dart/code.dart
+++ b/levenshtein/dart/code.dart
@@ -4,28 +4,28 @@ int levenshteinDistance(String word1, String word2) {
   final m = word1.length;
   final n = word2.length;
   // Create a matrix to store distances
-  var matrix = List.generate(m+1, (i) => List.generate(n+1, (j) => 0));
+  var matrix = List.generate(m + 1, (i) => List.generate(n + 1, (j) => 0));
 
-  // Initialize first row and column 
+  // Initialize first row and column
   for (int i = 1; i <= m; i++) {
-      matrix[i][0] = i;
+    matrix[i][0] = i;
   }
   for (int j = 1; j <= n; j++) {
-      matrix[0][j] = j;
+    matrix[0][j] = j;
   }
-  
+
   // Compute Levenshtein distance
   for (int i = 1; i <= m; i++) {
-      for (int j = 1; j <= n; j++) {
-        int cost = word1[i-1] == word2[j-1]? 0:1;
-        matrix[i][j] = min(
-            min(
-                matrix[i-1][j] + 1, // Deletion
-                matrix[i][j-1] + 1  // Insertion
-            ),
-            matrix[i-1][j-1] + cost // Substitution
-        );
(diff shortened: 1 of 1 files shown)


levenshtein/dart/code.dart
```diff
@@ -40,20 +47,20 @@ void main(List<String> args) {
...
   // Compare each pair of arguments exactly once
   for (int i = 0; i < args.length; i++) {
-      for (int j = i+1; j < args.length; j++) {
-        if (i != j) {
-            int distance = levenshteinDistance(args[i], args[j]);
-            if (minDistance == -1 || distance < minDistance ) {
-                minDistance = distance;
-            }
-            times++;
+    for (int j = i + 1; j < args.length; j++) {
```
PEZ, line 50: The Dart implementation fails the check for correctness:

```
❯ ../run.sh check  
...
Checking Dart
Incorrect output:
times: 1953
min_distance: 7
Check failed for Dart.
...
```

This is not introduced by your code, @vincevargadev. But you could include the fix anyway. Like so:

```dart
  // Compare each pair of arguments twice (i, j) and (j, i), to adhere to the reference implementations
  for (int i = 0; i < args.length; i++) {
    for (int j = 0; j < args.length; j++) {
```

Here the suggestion is to make it clear that we do the redundant check on purpose, so that people will not send PRs “fixing” it. 😄

PEZ (changes requested):
Thanks for helping in improving the project, @vincevargadev! 🙏 @bddicken and I did wonder why the Dart code was a bit slow, even though it skips half of the checks that the other languages do. Now we know!

Would be great if we made the Dart implementation benchmark compliant while we're at it. See code comments.

Ichoran:
This also doesn't use the C reference implementation algorithm which has two rows only.

PEZ:
Closing as we haven't heard back from @vincevargadev. May be relevant to revisit the ideas here when we have ported Dart over to the new runner. See:

* #371

PR #266, opened 2024-12-19
perf: reduce string comparisons in Levenshtein distance calculation

ramsyana:
Optimize Levenshtein distance algorithm performance

1. Reducing the total number of string comparisons by half
2. Making string character access more efficient 
3. Moving some calculations to compile time

Key changes:
- Changed string comparison loop to avoid duplicate checks
- Added character caching for faster access
- Used compile-time initialization for the first row

### DIFF
--- a/levenshtein/zig/code.zig
+++ b/levenshtein/zig/code.zig
@@ -21,8 +21,10 @@ fn levenshteinDistance(s1: []const u8, s2: []const u8) usize {
     var curr_row: [256]usize = undefined;
 
     // Initialize first row
-    for (0..m + 1) |i| {
-        prev_row[i] = i;
+    prev_row[0] = 0;
+    comptime var init: usize = 1;
+    inline while (init <= 255) : (init += 1) {
+        prev_row[init] = init;
     }
 
     // Main computation loop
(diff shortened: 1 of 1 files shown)


PEZ:
Hi! I don't think we are allowed to do **1** and **3**. 

The output of this version is
```
❯ ./zig/code `cat input.txt`                                   
times: 1953
min_distance: 7
```
Which is only half of the work the reference implementations do.

The benchmark tries to measure how fast a language does some work, and it's probably also cheating to do the work beforehand. (But @bddicken be the judge!)

Making string access more efficient should be perfectly fine, though. Do you know how much performance that unlocks?

ramsyana:
Hi @PEZ thanks. I've updated the code to increase the buffer size and optimize string comparisons by eliminating redundant pair checks. This should improve performance while handling longer strings.

Ichoran:
Part of the art of simple benchmarks is to find redundant busywork that you know is pointless but which the compiler can't discern is pointless.  So it makes sense to me to leave the redundant calculations in.  The benchmark already runs too fast for the results to be meaningful for languages that have a runtime engine with non-negligible startup time (e.g. JVM).

Gkodkod:
Hey @PEZ Not sure what you mean, as in my original code I had them and saw you experimented with taking them out. Not sure why. Both @ramsyana and @Ichoran points are valid. Agree with you that @bddicken should make the call.  Keep warm and safe. Have fun with your family during the holidays!

https://github.com/user-attachments/assets/2bf05006-3737-45c0-81ed-616403ef96af

PR #267, opened 2024-12-19
perf: improve performance of rust fibonacci code

danieloraca:
Improvements for the naive implementation:

Iterative Fibonacci:
* Avoids repeated recursion calls and achieves linear time complexity.

Graceful Error Handling:
* Instead of crashing on invalid input, the program provides meaningful error messages.

Range Fix:
* Changed the loop range to 0..u to include the 0th Fibonacci number.

Idiomatic Rust:
* Used iterators (map and sum) for concise and readable logic.

### DIFF
--- a/fibonacci/rust/code.rs
+++ b/fibonacci/rust/code.rs
@@ -1,20 +1,21 @@
 fn fibonacci(n: u32) -> u32 {
-    match n {
-        0 => 0,
-        1 => 1,
-        _ => fibonacci(n - 1) + fibonacci(n - 2),
+    let mut a = 0;
+    let mut b = 1;
+    for _ in 0..n {
+        let temp = a;
+        a = b;
+        b = temp + b;
     }
+    a
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


PEZ:
Hi, thanks for improving the implementation.

Can you explain why the changes improve things, and also please provide benchmarks for the before and after so that we better understand what ”improve performance” means here?

Ichoran:
This PR changes the algorithm.  It is no longer naive, and thus does not meet the requirements of the benchmark.

danieloraca:
Thanks for reviewing my PR.

The previous implementation of the Fibonacci function used a naive recursive approach, which violates basic principles of efficient programming. While it works for small input values, its exponential time complexity makes it impractical for larger inputs.

This update replaces the recursive implementation with an iterative approach, significantly improving performance and adhering to efficient programming practices. For example, the iterative version avoids redundant computations and achieves linear time complexity (O(n)), making it suitable for larger datasets.

PEZ:
It's a naïve fibonacci test.

PR #270, opened 2024-12-19
Add basic check for program correctness

PEZ:
To make it easier to implement and review benchmark we can make a basic test for each that checks the output.

For this I have updated the `run.sh` script so that it's default behaviour is to first run a check for correctness, then only if it passes the benchmark will be run. The script also takes an optional argument `check` that will cause the script to only check, no benchmark.

The script now takes the default input to the programs from the file `input.txt`, not just for levenshtein, but for all scripts. It also has a `check.sh` script per benchmark.

Further, the script accepts any arguments, that will override `input.txt` and also cause the run to be without checking. Because the `check.sh` scripts are really stupid and hardcoded on the `input.txt` input.

Sample run of the *loops* benchmark, with some selection of languages enabled:

```
~/Projects/languages/loops $ 
❯ ../run.sh     

Checking C
Check passed
Benchmarking C
Benchmark 1: ./c/code 40
  Time (mean ± σ):     476.9 ms ±  12.9 ms    [User: 466.9 ms, System: 3.5 ms]
  Range (min … max):   463.9 ms … 489.7 ms    3 runs
 

Checking Node
Check passed
Benchmarking Node
Benchmark 1: node ./js/code.js 40
  Time (mean ± σ):     755.7 ms ±   6.1 ms    [User: 749.0 ms, System: 6.7 ms]
  Range (min … max):   751.3 ms … 762.6 ms    3 runs
 

No executable or script found for Java. Skipping.

Checking Clojure
Check passed
Benchmarking Clojure
Benchmark 1: java -cp clojure/classes:src:/Users/pez/.m2/repository/org/clojure/clojure/1.12.0/cloju
  Time (mean ± σ):     754.7 ms ±   1.0 ms    [User: 987.3 ms, System: 58.6 ms]
  Range (min … max):   754.0 ms … 755.8 ms    3 runs
 

Checking Clojure Native
Check passed
Benchmarking Clojure Native
Benchmark 1: ./clojure-native-image/code 40
  Time (mean ± σ):     480.6 ms ±   0.3 ms    [User: 475.5 ms, System: 3.9 ms]
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.  Range (min … max):   480.4 ms … 480.9 ms    3 runs

 

Checking Bad input to Clojure Native
Output out of range 1950000 - 1959999:
455151
Check failed for Bad input to Clojure Native.

Checking Java GraalVM
Check passed
Benchmarking Java GraalVM
Benchmark 1: ./jvm.code 40
  Time (mean ± σ):     478.3 ms ±   3.5 ms    [User: 473.4 ms, System: 3.7 ms]
  Range (min … max):   474.4 ms … 481.1 ms    3 runs
```


bddicken:
This is great!

PR #271, opened 2024-12-20
Add scala-js with bun

He-Pin:
Motivation:
Scala-js has applied much more optimization than the scala-native one, so adding it to the bench metrix would be nice.

Modification:
Run the scala-js result with bun, as node is slower.

Result:
Test Scala-js with Bun


guizmaii (approved):
❤️

He-Pin:
@bddicken I think this one is ready.

PEZ:
Thanks! 🙏

Gkodkod:
This is awesome @He-Pin and thanks !!!

He-Pin:
@Gkodkod Based on your excellent work:)

Gkodkod:
@He-Pin  Thank you so much, and I appreciate your insights! 🙇

image

You mentioned: "The Scala-js and Scala-native did more optimization in the backend; the code is right, but the Scala 3 compiler currently lacks this optimization :(_" here https://github.com/bddicken/languages/pull/253#issuecomment-2558531599

One of my favorite Chinese saying is "千里之行，始于足下" (A journey of a thousand miles begins with a single step.) by Laozi
Another old Chinese saying goes, "One step backward, two steps forward"—though Scala 3 might not have caught up yet, every challenge is a step towards innovation.

Take a look at these optimization ideas and compiler enhancements for Scala 3 and let us know if we are on the right path.

In the absence of compiler-level optimizations, we as developers can refine our code to avoid unnecessary tuple allocations. Instead of creating a tuple for variable swapping, direct assignments can achieve the same logic without the overhead.

Example: Instead of this...

`val (str1, str2) = if (s1.length > s2.length) (s2, s1) else (s1, s2)`

We can refactor to:

```
val isS1Bigger = s1.length > s2.length
val str1 = if (isS1Bigger) s2 else s1
val str2 = if (isS1Bigger) s1 else s2
```

This refactoring avoids the creation of an intermediate tuple, thereby reducing unnecessary allocations and improving performance. 

Introducing a new phase in the Scala 3 compiler to specialize tuples, similar to how functions are specialized, could mitigate this issue. This approach is discussed in the Scala community as a way to enhance performance by reducing unnecessary allocations. See here on Scala 3 github 
https://github.com/scala/scala3/issues/11114
This one is more detailed on Scala 3 Performance metrrics is still open :( but with lots of insights
https://github.com/pityka/saddle/issues/428

If you're interested, there's a more detailed deep dive of Scala 3's New Compile-Time Features here, which might bring us more lightening ideas:
https://www.youtube.com/watch?time_continue=33&v=5PQYC0UDqGA&source_ve_path=MzY4NDIsMjg2NjY

Let us all know how we can improve it further, and thanks again for all your contributions! I'm looking into Spark caching optimizations now with Palantir in production, so wish me luck. 🙏
 
Additionally, with the holiday spirits, I have not gotten a chance to look at all the new optimizations and PR's, which I'm so delighted are taking place... Awesome progress, @bddicken and @PEZ, and certainly a further indication of how important this brillant idea iis.Happy New Year !!!

He-Pin:
Merry Christmas and best wishes too, all these is some kind of symphony

He-Pin:
@Gkodkod @PEZ 
There is an ASM bytecode for this task:
```java
public void swap(Type prev,
 Type type)
Generates the instructions to swap the top two stack values.
```

Not sure when Scala will implement this.

Gkodkod:
Super and Happy New Years @He-Pin !!!

He-Pin:
@Gkodkod Happy new year too

PR #273, opened 2024-12-20
feat: compile inner loop in O(1)

titaniumtraveler:
Because you can just do that if you know math!

Also seems like you are missing an `input.txt` for `loops`.
So currently you are benchmarking how fast different languages can unroll their
stack.

### DIFF
--- a/loops/rust/code.rs
+++ b/loops/rust/code.rs
@@ -10,10 +10,12 @@ fn main() {
         
     let mut a = [0u32; 10000];    // Array of 10k elements initialized to 0
     for i in 0..10000 {                  // 10k outer loop iterations
-        for j in 0..100000 {               // 100k inner loop iterations, per outer loop iteration
-            a[i] = a[i] + j % n;                // Simple sum
-        }
-        a[i] += r;                              // Add a random value to each element in array
+        let d = (100_000 - 1) / n;
+        let m = (100_000 - 1) % n;
+
+        a[i] += d * (n - 1) * n / 2 + m * (m + 1) / 2;
+
+        a[i] += r; // Add a random value to each element in array
     }
     println!("{}", a[r as usize]);              // Print out a single element from the array
 }


spytheman:
By that logic, all solutions can just print the results directly, if they know them ...

titaniumtraveler:
> By that logic, all solutions can just print the results directly, if they know them ...

I mean a lot of the inputs are *variable*.
But yes, the rest can be optimized out.

And I mean this specific optimization is mechanical enough that you could even teach it to a compiler.

necrophcodr:
> By that logic, all solutions can just print the results directly, if they know them ...

Surely a very VERY smart compiler would also optimize the inner loop as such, since it... Can.

PEZ:
I don't think this is in the spirit of this benchmark. It's trying to benchmark languages, not developers. When that smart compiler shows up, then sure.

necrophcodr:
> I don't think this is in the spirit of this benchmark. It's trying to benchmark languages, not developers. When that smart compiler shows up, then sure.

why then, is it not comparing languages equally? some languages might perform such optimizations, some might perform JITing in the right circumstances, and some might uses doubles or floats for their modulus operations which are faster than doing integer divisions. so is it comparing the languages 1:1? Or is it comparing how good the benchmark CPU is at handling the various types of division operations that languages provides? Or unrolling loops?

I don't mean to derail this conversation of course, but it is curious that this is not accounted for, so I'm not quite sure what the benchmarks are actually measuring, if a PR such as this would not fit the mark.

Additionally, try replacing the C example of using ints with doubles instead for the modulus operation, and suddently this example turns from being a slower-than-zig example to beating everything by miles:

```
➜ clang -O3 -march=native -mavx2 -lm fastc.c -o fast
➜ time ./slow 5
205825
./slow 5  1,45s user 0,00s system 99% cpu 1,455 total
➜ clang -O3 -march=native -mavx2 -lm slowc.c -o slow
➜ time ./fast 5
205612
./fast 5  0,28s user 0,00s system 99% cpu 0,283 total
```

So if it's a language benchmark, what does running these loops with integers infer, if other languages might use doubles? Maybe some of them in the future will be performing vectorized math and using SIMD to get better performance (like Mojo?)

It just seems odd, is all.

PEZ:
> why then, is it not comparing languages equally?

It's extremely hard to device benchmarks that manages to compare all languages equally/fair. There's simple an enormous diversity among languages. There already exists benchmarks that are quite rigorously defined. This benchmark project is taking a simpler, much less rigorous approach, to keep the maintenance burden reasonable.

Taking JIT into account would increase the complexity of the benchmark runner, and would make maintenance more difficult and time consuming. For some languages there exists ways to compile them into an optimized binary where the JIT has been given a chance to optimize. So at least for those languages there is a way to compete on more equal terms.

Of course we can discuss if there are clarifications we can add to the benchmarks. The question  about ints vs doubles seems like an area that could benefit from some light being shed.

PR #276, opened 2024-12-21
Add PR template

PEZ:
A tiny layer of bureaucracy. I see that a lot of the PRs lack description, and some have a bit too little of a description, or assume that everyone know some language at some expert level. Also, a description or title can have blanket statements like “this is faster”, without providing benchmark runs of before and after.


PEZ:
I merged without waiting for your feedback, @bddicken. Because the PRs are pouring in at a steady stream. 😄 We can remove or reword the template later if it is not to your liking.

PR #277, opened 2024-12-21
chore: fix scala compiles

He-Pin:
Motivation:
When run ../compile.sh ,if the name is not `scala` it prints nothing :(


He-Pin:
@PEZ Sorry for the inconvenience.

PEZ:
No worries. Thanks for fixing!

PR #278, opened 2024-12-21
Slightly optimize the odin levenshetin code

Waqar144:
See individual commits. First commit has changes, second commit just fixes the trailing spaces

### DIFF
--- a/levenshtein/odin/code.odin
+++ b/levenshtein/odin/code.odin
@@ -3,12 +3,12 @@ package main
 import "core:fmt"
 import "core:os"
 import "core:strings"
-import "core:runtime"
+import "base:runtime"
 
 // Calculates the Levenshtein distance between two strings using Wagner-Fischer algorithm
 // Space Complexity: O(min(m,n)) - only uses two arrays instead of full matrix
 // Time Complexity: O(m*n) where m and n are the lengths of the input strings
-levenshtein_distance :: proc(s1, s2: string) -> int {
+levenshtein_distance :: proc(s1, s2: string) -> int #no_bounds_check {
     // Early termination checks
     if s1 == s2 do return 0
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks! It would be great with the benchmark results from before and after the change. I don't know Odin, and can't judge “slightly optimize” from the code alone.

Waqar144:
> Thanks! It would be great with the benchmark results from before and after the change. I don't know Odin, and can't judge “slightly optimize” from the code alone.

The commit message has details which should make it easier to judge the changes:
```
- Fix warning, use base:runtime
- use builtin swap mechanism
- use temp allocator
- disable bounds checking for the function as we are always in bounds
```

I can post exact benchmark results later, but it should perform roughly equivalent to the C version now (or at least it does on my laptop)

Waqar144:
> Thanks! It would be great with the benchmark results from before and after the change. I don't know Odin, and can't judge “slightly optimize” from the code alone.

```
Checking Odin
Check passed
Benchmarking Odin
Benchmark 1: ./odin/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):      39.7 ms ±   2.3 ms    [User: 37.4 ms, System: 2.1 ms]
  Range (min … max):    37.1 ms …  41.2 ms    3 runs
 

Checking C
Check passed
Benchmarking C
  Time (mean ± σ):      42.1 ms ±   2.8 ms    [User: 40.5 ms, System: 1.1 ms]
  Range (min … max):    38.9 ms …  43.9 ms    3 runs

```

PEZ:
Thanks! To really help me judge the improvement, I would have appreciated runs from before the changes too, but I can see on the comparison with C that it is a real improvement, so I'm good with that.

PR #280, opened 2024-12-21
 hello-world: add Modula-2 implementation

stalecu:
There's not much going on here, using GNU Modula-2 (gm2). Since I don't have a Mac machine, I am not sure what to do for installing GCC and adding such a big dependency should be only worth it if we can also use GCC across the board for C and C++. 

Sidenote: according to [this Phoronix article](https://www.phoronix.com/news/COBOL-For-GCC-Patches-EOY2024), in 2025 maybe this repo can remove the GNU COBOL cask as well. There'll be probably another 20 languages added here by the time that happens, but this is just FYI.


PEZ:
Thanks! 🙏

stalecu:
@PEZ I noticed you merged my commit, but my question about GCC still stands. Does this repo already pull GCC? Because otherwise gm2 won't work (obviously), but GCC is big and as such we'd need to amortize the cost somewhat by moving to GCC for C and C++ (and perhaps Fortran as well... not sure about Objective-C).

PR #281, opened 2024-12-21
chore: support scala-js and scala the same time

He-Pin:
Motivation:
It was not good testing, becase I was using  a ARM laptop, after swtich to Windows, I find some problem.

Modification:
1. the scala js one when compiles to js and run with bun, the first two argumetns is not what we want, I have to adjust for that
2. for supporting scala 、scala-native and scalajs the sametime, I have to use the multi files script features
https://scala-cli.virtuslab.org/docs/guides/scripting/scripts#define-source-files-in-using-directives
,so I splited the files.

@PEZ Sorry again, but it works on x86 now.

refs: https://github.com/VirtusLab/scala-cli/issues/3391


levenshtein/scala/codeJS.scala
```diff
@@ -0,0 +1,18 @@
...
+@js.native
+@JSGlobal("process")
+object NodeProcess extends js.Object {
+  val argv: js.Array[String] = js.native
+}
+
+object MainJS {
+  def main(args: Array[String]): Unit = {
+    val params = NodeProcess.argv.drop(2)
+    LevenshteinDistance.run(params.toArray)
```
He-Pin, line 16: I have to get the arguments like this, @sjrd cc

He-Pin:
@PEZ Thanks

PEZ:
You think you can add Scala toolchains to the CI build, @He-Pin? That way we can see that things compile and can run hello-world.

He-Pin:
I think I can do that. but it's 4 am here, I will take a look tomorrow.

PEZ:
Oh, there's no hurry at all. The CI thing is new, lacking most toolchains, and we're not really depending on it yet. (It may actually prove not to scale, since there are so many toolchains that need to be there, and it already creates a huge cache, but we'll see.)

PR #283, opened 2024-12-22
Refactor Python Fibonacci

LordOdin:
Remove a redundant if statement from the python code.

Local testing showed ~60ms faster at 32 iterations.

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -2,10 +2,8 @@ import sys
 
 
 def fibonacci(n):
-    if n == 0:
-        return 0
-    if n == 1:
-        return 1
+    if n < 2:
+        return n
     return fibonacci(n-1) + fibonacci(n-2)
 
 


PEZ:
Hi! Thanks for contributing! 🙏 

At the moment the algorithm we're benchmarking is with the redundant checks. See this review comment as an example:

* https://github.com/bddicken/languages/pull/223#discussion_r1884765403

LordOdin:
> Hi! Thanks for contributing! 🙏
> 
> At the moment the algorithm we're benchmarking is with the redundant checks. See this review comment as an example:
> 
> * [Babashka: Use separate code from Clojure #223 (comment)](https://github.com/bddicken/languages/pull/223#discussion_r1884765403)

In that case checkout the D lang Fibonacci, it still has a single if

PEZ:
Yes, there are probably a few more languages where this has slipped through.

PR #285, opened 2024-12-22
Fixed The C Loop So you don't have to.

JustForCollege:
Added a casey inspired macro. 

Thanks,

### DIFF
--- a/loops/c/code.c
+++ b/loops/c/code.c
@@ -2,6 +2,7 @@
 #include "stdlib.h"
 #include "stdint.h"
 #include "time.h"
+#include "math.h"
 
 // A casey inspired macro
 #define REM(a, b) (int)(((double)a) - ((double)b)*trunc(((double)a) / ((double)b)))
@@ -13,7 +14,7 @@ int main (int argc, char** argv) {
   int32_t a[10000] = {0};              // Array of 10k elements initialized to 0
   for (int i = 0; i < 10000; i++) {    // 10k outer loop iterations
     for (int j = 0; j < 100000; j++) { // 100k inner loop iterations, per outer loop iteration
-      a[i] = a[i] + REM(j,u)               // Simple sum
+      a[i] = a[i] + REM(j,u);               // Simple sum
     }
     a[i] += r;                         // Add a random value to each element in array
   }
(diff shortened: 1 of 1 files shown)


loops/c/code.c
```diff
@@ -3,14 +3,17 @@
...
+
 int main (int argc, char** argv) {
   int u = atoi(argv[1]);               // Get an input number from the command line
   srand(time(NULL));                   // FIX random seed
   int r = rand() % 10000;              // Get a random integer 0 <= r < 10k
   int32_t a[10000] = {0};              // Array of 10k elements initialized to 0
   for (int i = 0; i < 10000; i++) {    // 10k outer loop iterations
     for (int j = 0; j < 100000; j++) { // 100k inner loop iterations, per outer loop iteration
-      a[i] = a[i] + j%u;               // Simple sum
+      a[i] = a[i] + REM(j,u)               // Simple sum
```
He-Pin, line 16: COOL, but why the compiler does not do

PEZ (commented):
Thanks for contributing! 🙏 

Can you explain for someone not familiar with Casey? What are the trade-offs? (I am assuming there are trade-offs since the compiler doesn't do this optimization.) And please also provide benchmarks showing what effects the change has to motivate the use of a macro.

JustForCollege:
@PEZ @He-Pin full explanation can be found in this video (https://www.youtube.com/watch?v=RrHGX1wwSYM), thing is the original code can't be vectorized, it uses idiv instruction (for example on x86 machines) which is very slow, compared to this version also uses division however this version the compiler can vectorize it, which is SIMD (Single Instruction Multiple Data), instead of working on one piece of data one instruction can manipulate multiple data, why the compiler doesn't do that already is out of my scope to be honest.

loops/c/code.c
JustForCollege: I don't know to be honest, the only time I saw a compiler optimizes modules is if you tried doing module power of 2, in that case it replaces it with "and" instruction, so for example if 

a = 35
b = 32 

a % b = a & (b - 1)  

which in this case is 3

PEZ:
Closing for now. This is for the legacy loops implementation. Open a new PR targeting the current implementation, if it is still relevant.

PR #288, opened 2024-12-22
Optimization for cpp impl of levenshtein

AndrewB330:
You do not need to actually swap all the values between two arrays, you can just swap "pointer" or "index" of the `prev` and `curr` arrays, so instead of having two separate arrays for two rows we can have an array or arrays of length 2 and use indexes to choose between prev and cur. In this case `(i&1)` is the index of the current row and `(i&1)^1` is the index of previous row on the step `i`. 

As you can notice on the `i+1` step the "prev" index `((i+1)&1)^1` will be equal to `(i&1)` or "curr" index on the step `i`.

P.S. Same optimization can be made for C implementation as well, but I don't have a setup to compile C code from this repository as it uses a version that allows non-constant size arrays on stack.

P.S.S most likely same optimization can be made for most of implementations here, but it depends a lot on language and maybe it will not improve things for some languages.

### DIFF
--- a/levenshtein/cpp/code.cpp
+++ b/levenshtein/cpp/code.cpp
@@ -81,7 +81,7 @@ int levenshtein(const string_view& str1, const string_view& str2) {
                 rows[prev][j] + 1,        // deletion
                 rows[curr][j - 1] + 1,    // insertion
                 rows[prev][j - 1] + cost  // substitution
-                });
+            });
         }
     }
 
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks for contributing! ❤️ 

@Gkodkod , the change is a bit opaque to me. Can you weigh in on wether it still adheres to the changes you made to the reference implementations?

Ichoran:
I don't think I like this optimization: it switches, fundamentally, from the two rows being their own variable, to the two rows being determined by index.

Furthermore, the way it does this changes the natural memory layout from aaaaa bbbbb to ab ab ab ab ab.  If a language provides this sort of reordering transparently as "best practice", fine.  But otherwise to make this switch would just require a ton of busywork for all the other language implementations if it turns out that the latter optimizes better for them.

So I'd judge this out of bounds as not-true-enough-to-the-reference-implementation.

PR #290, opened 2024-12-22
Improve Julia Levenshtein performance

jakobnissen:
This adds the following optimizations:
* Re-use the arrays instead of creating new ones in the levenshtein function, as allowed by the rules of the levenshtein README.md
* Operate on codeunits instead of strings. This means that the string is being interpreted as a byte array instead of a collection of UTF8 unicode characters. This will give incorrect answer for non-ASCII unicode strings, but this behaviour mirrors the C implementation
* Use 32-bit integers instead of 64-bit integers. This means the results may be wrong for very long strings, but this behaviour mirror the C impl also.
* Use the more low-level Memory instead of Vector for arrays

All in all, this results in an approximately 50% performance increase, from 26ms to 14 ms on my machine.
Of course, this handful of milliseconds is dwarfed by Julia's startup time and so won't make a dent in the actual timings, but I still think it's move appropriate to have the Julia implementation itself be fast.

Thanks and merry Christmas!

### DIFF
--- a/levenshtein/julia/code.jl
+++ b/levenshtein/julia/code.jl
@@ -1,9 +1,15 @@
 # Calculates the Levenshtein distance between two strings using Wagner-Fischer algorithm
 # Space Complexity: O(min(m,n)) - only uses two arrays instead of full matrix
 # Time Complexity: O(m*n) where m and n are the lengths of the input strings
-function levenshtein_distance(s1::String, s2::String)::Int
+function levenshtein_distance(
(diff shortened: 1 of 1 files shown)


PEZ:
Awesome. Thanks for helping with improving things here! 🙏 

We will hopefully find some way to give the languages with long start times a fairer treatment.

PEZ:

* #371
* #410

PR #293, opened 2024-12-23
chore: emit common js in scala-js

He-Pin:
I have:

## Description of changes

As 
image

It's now emit common js instead of the es module, two lines shorter.


He-Pin:
@PEZ I think this one is done.

PEZ:
Would still be cool with benchmarks before and after to see if it unlocked any performance. 😄

PR #295, opened 2024-12-27
 Add helloworld, loops and fibonacci in Emacs Lisp

a13:
I have:

## Description of changes
Added benchmarks in Emacs Lisp (somebody was going to do it anyway) :)

Not sure if I need to add anything Emacs related in Brewfile though.

Thanks


PEZ:
Very cool! I would like to compile the native version. But get complaints about Emacs not being compiled for native. If you have some tips how to fix this on a Mac for a total Emacs noob, I would be super happy! 😄

a13:
@PEZ 
Not sure if this is the best way, but this is what most Mac people use (I believe, since I don't have one, sorry)
```shell
brew tap d12frosted/emacs-plus
brew install emacs-plus@30 --with-native-comp
```

PEZ:
Thanks! It took a lot of time, but works.

Another thing that took a lot of time, before I installed the native Emacs, was to run the benchmarks for the byte code versions. 😄  

I have a small project where I visualize the results for the languages where I have a toolchain and patience to wait. Here's a permalink for the first and only appearance of Emacs Lisp Bytecode:

* https://pez.github.io/languages-visualizations/v2024.12.30/

Even slower than Python! Haha.

The native compilation is much faster and I will publish the visualizations including that very soon, and probably will let it stay on the charts.

PEZ:
Now with Emacs Lisp Native: https://pez.github.io/languages-visualizations/

PEZ:
@a13 , please consider contributing a port of this to the new runner:

* #371 

I just added Racket. Would be cool to get all them Lisps in. 😄

PR #318, opened 2025-01-02
Java: Enable native-image PGO compile

PEZ:
I have:

## Description of changes

The GraalVM native image compiles can be notched up a level by enabling PGO.

It improves Java's performance across the board. Here attaching **fibonacci** results from my M4 Mac.

## Fibonacci Before

```
Checking Java Native
Check passed
Benchmark 1: ./java-native-image/code 40
Benchmarking Java Native
Benchmark 1: ./java-native-image/code 40
  Time (mean ± σ):     288.6 ms ±   3.2 ms    [User: 282.6 ms, System: 3.7 ms]
  Range (min … max):   284.0 ms … 293.5 ms    7 runs
````

## Fibonacci After

```
Checking Java Native
Check passed
Benchmark 1: ./java-native-image/code 40
Benchmarking Java Native
Benchmark 1: ./java-native-image/code 40
  Time (mean ± σ):     263.3 ms ±   1.8 ms    [User: 258.0 ms, System: 3.4 ms]
  Range (min … max):   260.0 ms … 265.6 ms    7 runs
```

### CODE REVIEW

**Comment by Gkodkod** [comment]
@PEZ Excited to catch up next Sunday and looking forward to having more fun face-to-face with @bddicken! I'm a bit under the weather at the moment but hope to feel better by then. In the meantime, I’ll leave some comments here.

I’m also catching up on over 140 new updates from others (just fetched)—it’s heartwarming to see so much progress! I also realized I have some code on my branch that I forgot to check in for your review.

I absolutely love what you’ve done so far—you’ve practically "read my mind" regarding Java. I saved your optimized Clojure code and have been reflecting on it further.

As for Native and PGO, I’ve come up with this idea (apologies, my bed is calling, so I couldn’t prepare a full PR). Perhaps you could give it a try? I’m curious if it could further improve our results:

```
package jvm;

import java.util.concurrent.TimeUnit;
import jdk.incubator.foreign.MemorySegment;
import jdk.incubator.foreign.ValueLayout;

/**
 * Optimization Techniques:
 * Used @CompilerHint annotation to provide optimization hints
 * Added warm-up phase to help JIT compiler
 * Included performance timing
 * Maintained original recursive approach
 */

public class code {
    // Inline-friendly recursive Fibonacci implementation
    private static int fibonacci(int n) {
        if (n <= 1) return n;
        return fibonacci(n - 1) + fibonacci(n - 2);
    }

    // JIT and Native Image optimization hint
    @CompilerHint(inline = true)
    private static long computeFibonacciSum(int upperBound) {
        long result = 0;
        for (int i = 1; i < upperBound; i++) {
            result += fibonacci(i);
        }
        return result;
    }

    public static void main(String[] args) {
        // Input validation
        if (args.length == 0) {
            System.err.println("Please provide an upper bound as an argument.");
            System.exit(1);
        }

        try {
            int upperBound = Integer.parseInt(args[0]);
            
            // Warm-up phase for PGO
            for (int warmup = 0; warmup < 1000; warmup++) {
                fibonacci(10);
            }

            // Actual computation with timing
            long startTime = System.nanoTime();
            long result = computeFibonacciSum(upperBound);
            long endTime = System.nanoTime();

            System.out.println("Result: " + result);
            System.out.printf("Computation time: %.3f ms%n", 
                (endTime - startTime) / 1_000_000.0);
        } catch (NumberFormatException e) {
            System.err.println("Invalid input. Please provide a valid integer.");
            System.exit(1);
        }
    }

    // Annotation to help JIT and Native Image optimization
    @interface CompilerHint {
        boolean inline() default false;
    }
}
```

Bash Script - run.sh
```
#!/bin/bash
# PGO Native Image
# Run this script from the fibonacci/jvm directory
# Prerequisites: 
# - Install GraalVM
# - Set JAVA_HOME to GraalVM directory

# Compile Java source
javac --enable-preview optimized_code.java

# Generate profiling data
java -XX:+UnlockDiagnosticVMOptions \
     -XX:+TraceClassLoading \
     -XX:+LogCompilation \
     -XX:LogFile=fibonacci.log \
     optimized_code 30

# Create native image with PGO
native-image \
    --enable-preview \
    -H:Name=fibonacci \
    -H:+ReportExceptionStackTraces \
    --initialize-at-build-time \
    --no-fallback \
    optimized_code

# Run native executable
./optimized_code 30
```

Thank you for all the amazing progress and have a lovely weekend !

PEZ:
I think that as long as there are so big differences on how it behaves on two platforms that are so closely related, I think most changes are a bit like shots in the dark.

Gkodkod:
@PEZ  Thank you again !

1. The original code was written by Bruno Pacheco @brunopacheco1 
image

2. I am not so much into beating 1 language vs. the other and never been.  My main concern is to optimize performance in all languages if that is even possible, in the same manner.  Thanks to @He-Pin, I quickly also rewritten the Scala code to make it native.  See https://github.com/VirtusLab/scala-cli/issues/3401

3. You might want to try the run.sh script as it gives different args to the compiler and have not seen any optimized java code for fibonicci aside from above.  I am not sure exactly what you mean in your comments above referencing Readme and [benchmark descriptions](https://github.com/bddicken/languages/README.md#available-benchmarks) as my browser says Links not found.  

4. Maybe I missed something, but did not see soemthing similar to the run.sh script above so cannot really say anything about dark vs. light.  I prefer staying optimistic even when not feeling too well.

Lastly, I got to give some thought about the Static class you mentioned, and if you can expand or provide me a link to read that person's comment, will be able to perhaps learn something else.  Have a warm weekend buddy !

PR #323, opened 2025-01-05
Optimize zig

Kayden1412:
I have:

## Description of changes

Levenshtein: Use arena allocator.
Hello World: change print to writeAll
loop: use biased rng

### DIFF
--- a/levenshtein/zig/code.zig
+++ b/levenshtein/zig/code.zig
@@ -26,38 +26,34 @@ fn levenshteinDistance(s1: []const u8, s2: []const u8) usize {
     }
 
     // Main computation loop
-    var j: usize = 1;
-    while (j <= n) : (j += 1) {
+
+    for (1..(n + 1)) |j| {
         curr_row[0] = j;
 
-        var i: usize = 1;
-        while (i <= m) : (i += 1) {
+        for (1..(m + 1)) |i| {
             const cost: usize = if (str1[i - 1] == str2[j - 1]) 0 else 1;
-            
+
             // Calculate minimum of three operations
-            curr_row[i] = @min(
-                @min(
-                    prev_row[i] + 1,      // deletion
-                    curr_row[i - 1] + 1,  // insertion
-                ),
-                prev_row[i - 1] + cost    // substitution
+            curr_row[i] = @min(@min(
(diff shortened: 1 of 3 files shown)


PEZ:
Thanks for contributing! 🙏 

Please provide benchmarks using `run.sh` (with all other languages disabled, if you like) for before and after the change.

levenshtein/zig/code.zig
```diff
@@ -26,38 +26,34 @@ fn levenshteinDistance(s1: []const u8, s2: []const u8) usize {
...
 
     return prev_row[m];
 }
 
 pub fn main() !void {
-    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
-    defer _ = gpa.deinit();
-    const allocator = gpa.allocator();
+    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
+    defer arena.deinit();
```
scottredig, line 53: This deinit is unnecessary work.  When an arena allocator's lifespan is the programs anyways, let the OS do the cleanup.

levenshtein/zig/code.zig
```diff
@@ -26,38 +26,34 @@ fn levenshteinDistance(s1: []const u8, s2: []const u8) usize {
...
-            
+
             // Calculate minimum of three operations
-            curr_row[i] = @min(
-                @min(
-                    prev_row[i] + 1,      // deletion
-                    curr_row[i - 1] + 1,  // insertion
-                ),
-                prev_row[i - 1] + cost    // substitution
+            curr_row[i] = @min(@min(
```
scottredig, line 37: min can take 3 arguments.

scottredig (changes requested):

PEZ:
I merged this because I have run it and can confirm that it improves things. For levenshtein it even makes it work (before this change it hung on larger levenshtein input).

@scottredig , since it seems @Kayden1412 is busy with other things. You think you can PR in the changes you suggested? My Zig skills are about zero on a scale.

levenshtein/zig/code.zig
D-Berg: I think its "fairer" if zig cleans up its memory since I think a lot of the other languages does it.

PR #326, opened 2025-01-08
Improved levenshtein Swift code

nielshoek:
I have:

* [X] Read the project [README](../README.md), including the [benchmark descriptions](../README.md#available-benchmarks)

## Description of changes
Swift's `String` gives you the possibility to access the underlying C array/string. This gives a null terminated array, hence the `dropLast()`. This improves the speed quite significantly (on M3 Pro):
- Before: approx. 48ms
- After: approx. 28ms

It is rather straightforward as the code will show. Question is whether it is an allowed optimization.


PEZ:
LGTM. Thanks! 🙏

PR #327, opened 2025-01-09
Fix fortran clipping strings

saftacatalinmihai:
fix fortran clipping strings

I have:


PEZ:
Thanks for quick response!

Though I think we must consider longer strings. The input I used for some of my test has 750 character long strings.

I think we should require contributions to handle arbitrary length strings. I think that's what the C version does (as well as a lot of other implementations). This may be a significant factor for how the compiler can align data.

PR #332, opened 2025-01-10
levenshtein zig: handle arbitrary long input strings

### LINKED ISSUE #329: Levenshtein zig: Out of bounds on buffer
## Problem
The zig version uses two buffers, `prev_row` and `curr_row` of length 256 in `fn levenshteinDistance` while the longest input word has length of 258 causing possible undefined behaviour. 

Also `@min` can take 3 arguments so there is no need to call it twice.

D-Berg:
I have:

## Description of changes

- replaced fixed buffers of length 256 with dynamic buffer of length of longest input string.
- make only one call to `@min` since it can take 2 or more args.
- Reuses the dynamically allocated buffers prev and curr.

This closes #329 and also implement the suggestions in #323.

### DIFF
--- a/levenshtein/zig/code.zig
+++ b/levenshtein/zig/code.zig
@@ -3,7 +3,7 @@ const std = @import("std");
 /// Calculates the Levenshtein distance between two strings using Wagner-Fischer algorithm
 /// Space Complexity: O(min(m,n)) - only uses two arrays instead of full matrix
 /// Time Complexity: O(m*n) where m and n are the lengths of the input strings
-fn levenshteinDistance(s1: []const u8, s2: []const u8) usize {
+fn levenshteinDistance(s1: []const u8, s2: []const u8, prev_row: []usize, curr_row: []usize) usize {
     // Early termination checks
     if (std.mem.eql(u8, s1, s2)) return 0;
     if (s1.len == 0) return s2.len;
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks! 🙏 

Seems your changes also fixed something with an unclean exit:

**Before**:

```
Checking Zig
Check passed
Benchmarking Zig
Benchmark 1: ./zig/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     595.0 ms ±   6.1 ms    [User: 574.7 ms, System: 4.8 ms]
  Range (min … max):   586.9 ms … 603.4 ms    7 runs
 
  Warning: Ignoring non-zero exit code.
```

**After**:

```
Checking Zig
Check passed
Benchmarking Zig
Benchmark 1: ./zig/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     589.3 ms ±   2.5 ms    [User: 578.3 ms, System: 4.9 ms]
  Range (min … max):   585.7 ms … 592.8 ms    7 runs
```

Note to self (and @bddicken): We should be able to pick up non-zero exit codes in the correctness check. Running benchmarks for programs not exiting cleanly doesn't make sense.

PR #335, opened 2025-01-10
Optimize rust code of loop and levenshtein

Paulo-21:
I have:

* [X] Read the project [README](../README.md), including the [benchmark descriptions](../README.md#available-benchmarks)

## Description of changes
Change the rust loop to avoid bound checking, this is an optimisation a way that the rust officiel book tell us when we start to learn rust so it's not a cheat.

## Loop
Rust check every time if your index is higher than the lenght of the vector for memory safety.
But if we iterate over the vector, the rust compiler can undestund that we want to iterat over the vector only and it will not go futher than the lenght of the vector and ovoid bound checking.

## Levenshtein
- Less useless branch
- Reduce the lenght of the vector who was set with a x2 lenght and use int32 like the code instead of u64.

Loop : 2.0 sec -> 1.6 sec
levenshtein : 1.6 sec-> 1.0 sec

### DIFF
--- a/levenshtein/rust/code.rs
+++ b/levenshtein/rust/code.rs
@@ -31,15 +31,13 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
 
     // Use two rows instead of full matrix for space optimization
     let mut prev_row = Vec::with_capacity(m + 1);
-    let mut curr_row = Vec::with_capacity(m + 1);
-
+    for j in 0..=prev_row.capacity() {prev_row.push(j as i32);}
+    let mut curr_row:Vec<i32> = vec![0;m+1];
     // Initialize first row
-    prev_row.extend(0..=m);
-    curr_row.resize(m + 1, 0);
 
     // Main computation loop
     for j in 1..=n {
-        curr_row[0] = j;
+        curr_row[0] = j as i32;
 
         for i in 1..=m {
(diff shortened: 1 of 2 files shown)


PEZ:
Yay!

It looks good to me at a glance, but my Rust is not fluent at all. Please provide `run.sh` results for before and after the change, so that we easier can assess the changes.

Paulo-21:
## Loop
Before : 
image
Well i know it s wird, i m  not able to reproduce the previous time but it 's unexpected, maybe i was too tired when testing XD 
"After" :
image
Just for flex  : 
image
RUSTFLAGS='-C target-cpu=native' cargo build -r 
I saw g++ using native option so why do Rust doesn't use it ?
image

## Levenshtein
Before : 
image
After : 
image

zierf:
In [Commit #ba17989](https://github.com/bddicken/languages/pull/335/commits/ba1798905a1c77c5e0be1f0af3f159a522f2aee0) you replaced following code:
```Rust
                let distance = levenshtein_distance(&args[i], &args[j]);

                if let Some(current_min) = min_distance {
                    if distance < current_min {
                        min_distance = Some(distance);
                    }
                } else {
                    min_distance = Some(distance);
                }
```

Interestingly, the new variant with the constant minimum allocation runs _~11%_ slower.
```Rust
                let distance = levenshtein_distance(&args[i], &args[j]);
                min_distance = std::cmp::min(distance, min_distance);
```

PR #342, opened 2025-01-11
Update code.py

Draeli:
A Pythonic version of the code leads to more realistic results.

I have:

### DIFF
--- a/loops/py/code.py
+++ b/loops/py/code.py
@@ -1,16 +1,18 @@
 import sys
 import random
 
-
 def main():
     u = int(sys.argv[1])  # Get an input number from the command line
     r = random.randint(0, 10000)  # Get a random number 0 <= r < 10k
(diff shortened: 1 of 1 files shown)


mo-ba:
:100: :clap: 
this exposes the nonsensical implementation and irrelevance of the whole project.

you now have three choices

1. keep the slow for loop in python and have a test that is absolute BS for python because loops for loops are always a bad choice.
2. make the same "optimization" and un-nest the loops in all languages
3. we just close this issue and ignore the problems you just uncovered

PEZ:
Hi! This is not in line with wha the benchmark is about. From [loops/README.md](https://github.com/bddicken/languages/blob/main/loops/README.md):

> The code is designed to try force the compiler to generate code that actually makes 100 million updates.
> The code should be free of any hints to the compiler that the loops actually can be fully unrolled.

PR #343, opened 2025-01-11
Update code.py

Draeli:
Pythonic version and proper algorithm

I have:

* [ ] Read the project [README](../README.md), including the [benchmark descriptions](../README.md#available-benchmarks)

### DIFF
--- a/fibonacci/py/code.py
+++ b/fibonacci/py/code.py
@@ -1,20 +1,26 @@
 import sys
 
-
-def fibonacci(n):
+def fibonacci_iterative(n):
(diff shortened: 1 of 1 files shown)


PEZ:
Hi! This is not in line with the benchmark. From [fibonacci/README.md](https://github.com/bddicken/languages/blob/main/fibonacci/README.md):

> * For the non-base cases the code should do two recursive calls.
> * The code should be free of any hints to the compiler to memoize, use tail recursion, iterative methods, or any avoidance of the naïve recursion.

PR #346, opened 2025-01-12
Making Rust faster than C with this one simple trick (optimization for levenshtein)

tp971:
I have:

## Description of changes

Replacing the inclusive ranges `1..=n` and `1..=m` by exclusive ranges `1..n + 1` and `1..m + 1` in the Rust version of Levenshtein improves the performance by about 25% on my machine (making it about 3% faster than the C version). The reason for that is that the implementation of inclusive ranges accommodates for the fact that the upper bound might be the largest representable integer (e.g. `1..=usize::MAX`) by adding additional code that is executed in every iteration of the loop, which might also prevent certain compiler optimizations from running, making the performance even worse.

### DIFF
--- a/levenshtein/rust/code.rs
+++ b/levenshtein/rust/code.rs
@@ -38,10 +38,10 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
     curr_row.resize(m + 1, 0);
 
     // Main computation loop
-    for j in 1..=n {
+    for j in 1..n + 1 {
         curr_row[0] = j;
 
-        for i in 1..=m {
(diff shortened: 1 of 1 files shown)


zierf:
For me using `enumerate` on the string bytes instead of exclusive ranges saves _~10%_.

see also [PR #351](https://github.com/bddicken/languages/pull/351#issuecomment-2586786157)

PEZ:
Closing. Please check the new `levenshtein.rs` module and check if you think it can still be improved with the idea in this PR, @tp971 . Then a new PR is better than trying to retrofit this one, I think.

PR #348, opened 2025-01-12
Just adding an else to the if gets you about 30% performance

zveinn:
This can probably be optimized more, but I'm sleeping and want to watch some anime.

Not am expert on the algo, so maybe the else breaks it ? but the prints are the same


zveinn:
time == microseconds

# this is with else
```bash
 ./lev  
time: 839
times: 7656
min_distance: 3
min_distance: 796
!2049 sveinn lol
 go build .
```


# this is without else
```bash
 ./lev
time: 1157
times: 7656
min_distance: 3
max_distance: 796
!2051 sveinn lol
```

zveinn:
nvm, the else breaks the non min and max distances.. sadge

Issue #320, opened 2025-01-04
Restructuring compilation, benchmarking


benknoble:
As I mentioned in #319 and comments, I built a series of patches on top of ca443ca (update, 2024-12-17) over the holidays. The rapid progression since that commit made a simple rewrite or rebase a bit trickier, so I've put off doing it, but I'd like to discuss the restructuring a bit first and suggest some ways it could be handled with latest main.

## Compilation, Cleaning

The main idea is to move the compilation and clean commands to a POSIX-portable make syntax using suffix rules. Here's an example rule for a non-C language just to show it can be done:

```make
.rkt:
	raco make $<
	raco demod -o $*.zo $<
	raco exe $(RACO_FLAGS) -o $@ $*.zo
```

After adding `.rkt` to `.SUFFIXES`, then the idea is that `cd bench && make racket/code` would just do the right thing (in this case, compiling the file, demodularizing it, and packaging up a standalone executable). Further, artifacts could be listed for a `PHONY` `clean` target to remove, and all compiled programs would be dependencies of the (default by virtue of being first) `all` target, so that instead of `cd bench && ../compile.sh` you run `cd bench && make`.

The magic that makes this happen in the current series is putting a universal `base.make` in the root and then a simple shim in each benchmark:

```
.POSIX:
SHELL = /bin/sh

include ../base.make
```

## Benchmarking

There are less big changes here. Primarily:

- removing unneeded `.sh` prefixes (that is typical of libraries but not programs)
- not ignoring benchmark program failures (which should be a bug for the relevant program)
- throwing all commands into a single `hyperfine` run to get relative speed reports

I also tweaked hyperfine settings (the number of runs and warmups seemed low for statistics to me) and, against my better judgement, kept the `sed` command, which makes `hyperfine`s output less useful by disabling any indicators of progress.

Long term, it would be nice to be able to run a named subset of programs (my script limits programs by checking whether the program can be invoked) to allow iterating on a specific language with the repo's machinery rather than extra-judicially (_i.e._, running programs directly in the shell). This means that contributors wouldn't need to install all the languages and pay for all the compilation time if they didn't want to (with a Makefile, you can always opt out of compiling some targets by specifying only those you want to build).

---

[...]

PR #350, opened 2025-01-13
Rework hyperfine invocation

benknoble:
I have:

## Description of changes

convert run.sh to run

Importantly, it puts all the possible commands into a single hyperfine
invocation, which can help you get a relative comparison in the final
report (see below). Also, use a normal number of runs and warmups to get
more statistically significant results.

The "runnable" test also improves: it checks that the program exists &&
that the runner program (which may be the same) exists and is
executable. That means if you don't have `miniruby` installed, for
example, it should be skipped.

The use of Zsh is for portability with associative arrays (macOS's
/bin/bash is still too old on most systems).

Other noteworthy changes:
- dropped Hare because it didn't appear to work with existing script
- dropped commented out programs
- "descriptions" lost, but could be re-added in a separate parallel
  array
- kept Clojure program filename, though it doesn't look right to me
- common-lisp file check includes .lisp suffix (in run.sh, only the
  command had the suffix)
- Clojure runner construction with -Spath might not work…
- do _not_ ignore failures when running benchmarks (8d3432e (run: do _not_ ignore failures when running benchmarks, 2025-01-12))

Example run with just Racket and C:

    # ../run
    Benchmark 1: racket/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...
      Time (mean ± σ):      5.327 s ±  0.197 s    [User: 4.991 s, System: 0.112 s]
      Range (min … max):    5.118 s …  5.700 s    10 runs

    Benchmark 2: c/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...
      Time (mean ± σ):     836.5 ms ±   3.8 ms    [User: 832.3 ms, System: 1.9 ms]
      Range (min … max):   832.5 ms … 843.3 ms    10 runs

    Summary
      c/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa... ran
        6.37 ± 0.24 times faster than racket/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...

Makes-progress-on: https://github.com/bddicken/languages/issues/320

---

This may still be too big as a single PR (and is not how I'd prefer to present the changes); if it's helpful, I can try to split out most of the "other noteworthy changes" into separate commits (one of them already is). NB that takes more work on my end ;)


PEZ:
Hi! With #370 we no longer use hyperfine for the main benchmarking. 

I do think some of your “Other noteworthy changes” may still apply. I understand if you don't want to do all this work again, and if so, I will use your commits as a TODO list for myself. 😄

PR #353, opened 2025-01-13
Improve crystal levenshtein and make closer to c reference

Tamnac:
I have:

## Description of changes

Hi, same guy who originally added crystal. I noticed the levenshtein implementation was not entirely idiomatic and also had *terrible performance*, which is unexpected. This PR fixes that. Changes:
- Use built in initializer for array (more idiomatic)
- Use `to_slice` instead of `bytes`
- Don't allocate an array just to calculate the min (this was the reason behind the abysmal performance fyi, the other changes are relatively minor) 
- Use `each_permuation` instead of doing it by hand
- Swap i & j (cause who uses j first??)
- Don't check for math overflow (to be closer to C reference)

Also `StaticArray` probably should be a regular Array so it's not hardcoded 1024, but that's been already merged so I'll leave it.

P.S Crystal has a built in levenshtein distance module, that uses this same formula, but uses some `unsafe` stuff to run faster, which I think would be against the spirit of this benchmark.

### DIFF
--- a/levenshtein/crystal/code.cr
+++ b/levenshtein/crystal/code.cr
@@ -14,35 +14,31 @@ def levenshtein_distance(s1 : String, s2 : String) : Int32
   n = s2.size
 
   # Use two arrays instead of full matrix for space optimization
-  prev_row = StaticArray(Int32, 1024).new(0)
+  prev_row = StaticArray(Int32, 1024).new {|i| i}
   curr_row = StaticArray(Int32, 1024).new(0)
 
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks! I'm sorry for the delay.

FYI: The project is moving over to an in-process benchmarking method, where we try to measure as little as possible of things outside the actual benchmark. If you want to help with putting Crystal on those wheels, please see the project README and issue #371 .

PR #354, opened 2025-01-13
Zig arrays to be stack-allocated

PDeveloper:
I have:

## Description of changes

To be closer to the C or Fortran versions, use stack-allocated arrays in the Zig version instead of dynamically heap-allocated arrays, providing more accurate results.

### DIFF
--- a/levenshtein/zig/code.zig
+++ b/levenshtein/zig/code.zig
@@ -3,7 +3,7 @@ const std = @import("std");
 /// Calculates the Levenshtein distance between two strings using Wagner-Fischer algorithm
 /// Space Complexity: O(min(m,n)) - only uses two arrays instead of full matrix
 /// Time Complexity: O(m*n) where m and n are the lengths of the input strings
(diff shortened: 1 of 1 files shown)


levenshtein/zig/code.zig
```diff
@@ -16,6 +15,10 @@ fn levenshteinDistance(s1: []const u8, s2: []const u8, prev_row: []usize, curr_r
     const m = str1.len;
     const n = str2.len;
 
+    // Initialize arrays
+    var prev_row: [1024]usize = undefined;
+    var curr_row: [1024]usize = undefined;
```
Waqar144, line 20: this is broken code and will panic with some inputs. C is doing VLAs which is also broken (e.g,. very large strings that overflow the stack).

D-Berg:
See this [comment](https://github.com/bddicken/languages/issues/329#issuecomment-2581311488) on why the arrays are heap allocated.

zierf:
In order to be able to use arbitrary input sizes, stack-allocated arrays at comptime with fixed sizes are no longer an option.

PR #355, opened 2025-01-13
set row length to string length plus one

zierf:
Fixes following error:
`panic: index out of bounds: index 750, len 750`

---

I have:

## Description of changes

This error can be recreated by omitting the `ReleaseFast` flag for compilation or using `ReleaseSafe` instead.
```Zig
zig build-exe -O ReleaseSafe …
```

This way Zig will always check array bounds when accessing elements via their index and throw a panic.
The maximum number of elements for the rows only contain `string length` elements, but their capacity should be `string length + 1`.

The [_ReleaseFast_ flag omits all safety checks](https://ziglang.org/documentation/master/#ReleaseFast) and hides the underlying problem.

---

Apart from that, the Zig code takes almost twice as long as the C(pp) and Rust codes. Unfortunately, I am not familiar enough with Zig to find the exact cause.

### DIFF
--- a/levenshtein/zig/code.zig
+++ b/levenshtein/zig/code.zig
@@ -64,8 +64,8 @@ pub fn main() !void {
     const max_inp_len = std.mem.max(usize, input_lengths);
 
     // Reuse prev and curr row to minimize allocations
-    const prev_row = try allocator.alloc(usize, max_inp_len);
-    const curr_row = try allocator.alloc(usize, max_inp_len);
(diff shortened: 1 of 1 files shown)


zierf:
The issue addressed in this pull request is also fixed in [PR #373](https://github.com/bddicken/languages/pull/373) through equivalent adjustments.

zierf:
Replaced by [PR #373](https://github.com/bddicken/languages/pull/373)

PR #357, opened 2025-01-13
Make C# Levenshtein faster

janosorcsik:
I have:

## Description of changes

- Use swapping instead of a temp variable to replace array values.
- Use regular arrays because they're already initialized with 0.
- Only check the length of strings to remove extra branching cause by string.IsNullOrEmpty method


AnonymousRetard:
Have you actually tested that this is faster?
Reading your changes I'm pretty sure a lot of them is actually making the code slower.
By creating arrays using "new" you are now causing heap allocations which are slower than stack allocated memory and require garbage collection.

Comparing strings with "==" is not fast at all, it first checks if either string is null before calling string.Equals.
Changing from "int" to "var" shouldn't make any difference to performance at all.

Probably largest time taken by the C# is just startup time when the IL gets compiled to machine code. If code was compiled with NativeAOT it should run much faster.

PEZ:
NativeAOT does seem to loose out a tiny bit with the run times of the new input data. It just barely runs the test faster than the non-AOT version, even though it starts much faster:

* https://pez.github.io/languages-visualizations/#levenshtein
* https://pez.github.io/languages-visualizations/#hello-world

AhmedZero:
Test it using net 8.0 it'll be faster.
There is problem in net 9.0.

PEZ:
Would be awesome if someone wanted to help getting C# over to the new benchmark runner:

* https://github.com/bddicken/languages/issues/371

PR #360, opened 2025-01-14
Improve rust performace for levenshtein

PizzasBear:
I have:

## Description of changes

Rust performance seemed to be subpar so I improved it.

First of all, I replaces `usize`s (8 bytes on a 64-bit machine) with `u32`s (4 bytes) in the row vectors because it improves caching and while less correct I added the necessary checks so that it'd never silently fail.
I also added an ASCII check to make the byte conversion correct, because while other languages may ignore it, UTF-8 characters are not always one byte sized.
In addition I rewrote the loops so that there would be no out-of-bounds checks without altering the algorithm.
Finally, I rewrote portions of the code to make it more modern and idiomattic.

image

### DIFF
--- a/levenshtein/rust/code.rs
+++ b/levenshtein/rust/code.rs
@@ -7,58 +7,50 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
     // Early termination checks
     if s1 == s2 {
         return 0;
-    }
-    if s1.is_empty() {
+    } else if s1.is_empty() {
         return s2.len();
-    }
-    if s2.is_empty() {
+    } else if s2.is_empty() {
         return s1.len();
     }
 
-    // Convert to bytes for faster access
-    let s1_bytes = s1.as_bytes();
-    let s2_bytes = s2.as_bytes();
+    // Correctness checks
+    assert!(s1.len().max(s2.len()) < u32::MAX as usize / 2);
+    assert!(s1.is_ascii() && s2.is_ascii());
 
-    // Make s1 the shorter string for space optimization
-    let (s1_bytes, s2_bytes) = if s1_bytes.len() > s2_bytes.len() {
-        (s2_bytes, s1_bytes)
-    } else {
-        (s1_bytes, s2_bytes)
+    // Convert to bytes for faster access
+    // And make s1 the shorter string for space optimization
+    let (s1, s2) = match s1.len() < s2.len() {
+        true => (s1.as_bytes(), s2.as_bytes()),
+        false => (s2.as_bytes(), s1.as_bytes()),
     };
 
-    let m = s1_bytes.len();
-    let n = s2_bytes.len();
-
     // Use two rows instead of full matrix for space optimization
-    let mut prev_row = Vec::with_capacity(m + 1);
-    let mut curr_row = Vec::with_capacity(m + 1);
-
-    // Initialize first row
-    prev_row.extend(0..=m);
-    curr_row.resize(m + 1, 0);
+    let mut prev_row: Vec<_> = (0..=s1.len() as _).collect();
+    let mut curr_row = vec![0u32; s1.len() + 1];
 
     // Main computation loop
-    for j in 1..=n {
-        curr_row[0] = j;
+    for (j, &ch2) in s2.iter().enumerate() {
+        curr_row[0] = (j + 1) as _;
 
-        for i in 1..=m {
-            let cost = if s1_bytes[i - 1] == s2_bytes[j - 1] { 0 } else { 1 };
-            
+        // Iterate on 2-sized windows of curr_row and prev_row to remove out-of-bounds checks
+        let mut curr_window0 = curr_row[0];
+        for (&ch1, (curr_window1, prev_window)) in
+            (s1.iter()).zip(curr_row[1..].iter_mut().zip(prev_row.windows(2)))
+        {
+            let cost = (ch1 != ch2) as u32;
+
             // Calculate minimum of three operations
-            curr_row[i] = std::cmp::min(
-                std::cmp::min(
-                    prev_row[i] + 1,      // deletion
-                    curr_row[i - 1] + 1,  // insertion
-                ),
-                prev_row[i - 1] + cost    // substitution
-            );
+            *curr_window1 = (prev_window[1] + 1) // deletion
+                .min(curr_window0 + 1) // insertion
+                .min(prev_window[0] + cost); // substitution
+            curr_window0 = *curr_window1;
         }
 
         // Swap rows
-        std::mem::swap(&mut prev_row, &mut curr_row);
+        (prev_row, curr_row) = (curr_row, prev_row);
     }
 
-    prev_row[m]
+    *prev_row.last().unwrap() as _
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks! There's a PR for Rust Levenshtein performance and a discussion here:
* #351

Paulo-21:
Like this code btw, as an authors of one the rust improvement tentative for levenshtein in one of the PR's

levenshtein/rust/code.rs
```diff
@@ -7,58 +7,50 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
...
-        curr_row[0] = j;
+    for (j, &ch2) in s2.iter().enumerate() {
+        curr_row[0] = (j + 1) as _;
+
+        // Iterate on 2-sized windows of curr_row and prev_row to remove out-of-bounds checks
+        let mut curr_window0 = curr_row[0];
+        for (&ch1, (curr_window1, prev_window)) in
+            (s1.iter()).zip(curr_row[1..].iter_mut().zip(prev_row.windows(2)))
+        {
+            let cost = (ch1 != ch2) as u32;
```
zierf, line 40: Is it guaranteed that the boolean expression here is always converted to a "1"?
I also preferred the IF condition because it is easier to read.

If the condition can be optimized away by the compiler and the runtime is not increased, I would prefer to return a 0u32/1u32 similar to before.

levenshtein/rust/code.rs
```diff
@@ -7,58 +7,50 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
...
-    let n = s2_bytes.len();
-
     // Use two rows instead of full matrix for space optimization
-    let mut prev_row = Vec::with_capacity(m + 1);
-    let mut curr_row = Vec::with_capacity(m + 1);
-
-    // Initialize first row
-    prev_row.extend(0..=m);
-    curr_row.resize(m + 1, 0);
+    let mut prev_row: Vec<_> = (0..=s1.len() as _).collect();
```
zierf, line 28: For many, replacing `0..=m` with `0..m+1` had a small performance gain.
Using exclusive ranges may work here too.

levenshtein/rust/code.rs
Paulo-21: Yes it's a branchless version
I prefer the branchless version imo
I readability is your purpose we can add a comment ?

levenshtein/rust/code.rs
```diff
@@ -7,58 +7,50 @@ fn levenshtein_distance(s1: &str, s2: &str) -> usize {
...
+    let mut curr_row = vec![0u32; s1.len() + 1];
```
CptPotato, line 29: Wouldn't it make more sense to move these allocations out of `levenshtein_distance` and reuse the buffer instead?

Something along [these lines](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=efaf844f8e534d75309e18a9796642c0) (haven't tested thorougly)

PEZ:
Some changes made in #351 referenced changes made in this PR. Are there remaining changes that should still be considered?

levenshtein/rust/code.rs
zierf: Yes, the Zig code, for example, has a similar implementation, only with two (too small) buffers and I have created a [new variant with one larger buffer](https://github.com/bddicken/languages/pull/373/files#diff-cb854d404829d8b835c9ecf677804e57c5a7b0b5bb4df2a2ae01d43aab5bd644R77). However, its size is already determined when the variable is created so that no unnecessary resizes occur later. This also works for Rust, I have already tried it out in a local example with some minor improvements.

It's just questionable whether this should still be part of this PR.

zierf:
What I like better about this PR is that it [passes string slices](https://github.com/bddicken/languages/blob/74c1ff810abfd0b2d91f3ee495ef187be610f38f/levenshtein/rust/code.rs#L6) (`&str`) instead of string bytes (`&[u8]`) to the `levenshtein_distance` function. This feels more correct for a function that is supposed to compare strings.

The [calculation of the substitution costs also uses a casting from bool to u32](https://github.com/bddicken/languages/blob/74c1ff810abfd0b2d91f3ee495ef187be610f38f/levenshtein/rust/code.rs#L40) instead of an additional IF condition.

At the same time, I find that iterating over two windows is much harder to read than the current implementation. Unfortunately, I cannot judge whether merging the window iterators really saves any additional costs.

---

@omalley stated in [this comment](https://github.com/bddicken/languages/pull/351#issuecomment-2591398553) that swapping rows by reassigning would be faster (tested `M2 Max`).
`(prev_row, curr_row) = (curr_row, prev_row);`

However, @sammysheep has observed the opposite and preferred the swap function (tested `M4 Max`).
`std::mem::swap(&mut prev_row, &mut curr_row);`

It would be great to have another quick measurement for this. Maybe also from @PizzasBear themself with whatever machine/CPU. Using `std::mem::swap` or reassigning the variables doesn't make a difference to me.

---

I have run benchmarks where the only difference is in the _Main computation loop_. Taken from [PR #351](https://github.com/bddicken/languages/pull/351) and then also from this PR, running on a `AMD Ryzen 9 5950X 16-Core Processor`.

Benchmark with [PR #351](https://github.com/bddicken/languages/pull/351):
```SH
Benchmarking Rust
Benchmark 1: ./rust/target/release/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     493.4 ms ±   3.0 ms    [User: 490.5 ms, System: 1.3 ms]
  Range (min … max):   489.9 ms … 495.7 ms    3 runs
```

Benchmark with this PR:
```SH
Benchmarking Rust
Benchmark 1: ./rust/target/release/code aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ...
  Time (mean ± σ):     492.2 ms ±   2.6 ms    [User: 489.9 ms, System: 1.3 ms]
  Range (min … max):   489.7 ms … 494.9 ms    3 runs
```

For me, the runtimes for both variants are so close together at around 490-495 ms that there is no clear winner and everything is within the margin of error. In this case I also used `std::mem::swap` for both of my tests.

Perhaps Rust's Zero Cost Abstractions have already optimized this as much as possible. If someone can take this apart and compare the assembler codes, please do.
In this case I would prefer the readability of [PR #351](https://github.com/bddicken/languages/pull/351) for the [Main computation loop](https://github.com/sammysheep/languages/blob/5325d6d6f1dd9023f1b90f5f45b0ceaf950dec1d/levenshtein/rust/code.rs#L19-L36).

---

@PizzasBear could you apply the changes from the _main_ branch for this PR and add your preferred ones on top if you still notice an improvement?

Among other things, there was a change from `usize` to `u32`, which doesn't make much difference for me, but makes a big difference for Mac users.

sammysheep:
Point of order: `&str` is for UTF-8 strings but this function only handles ASCII. Rust does not have a first party ASCII string type, albeit `b"ascii"` can be used for byte strings in literals. One can write a Unicode levenshtein distance function using `chars()`, but this benchmark does not consider that for its canonical case. I chose byte slices as more representative of the behavior a caller can expect from the parameters.

PEZ:
Closing this. A lot of things have changed and I think it's better that we open new PRs for any remaining improvements. Let me know if you think this is a bad call, @PizzasBear, and we can reopen.

PR #362, opened 2025-01-14
CPP: Update C++ coding style; clarified code

int2str:
Since there are a lot of eyes on this code right now, the following changes showcase some C++ best practices and differentiate it a bit from the C solution.

No logic or performance changes were made to preserve the core algorithm.

- Sorted include statements Generally, a tool like clang-format does this for you.

- Removed "using namespace std;" "using namespace" directives are generally frowned upon. Google's coding style forbids them outright [1], the C++ Core Guidelines makes some exemptions, but also discourages them [2]. Even in this very small piece of code, this already caused a potential problem. The variable "prev" would shadow the std::prev() function template.

  1: https://google.github.io/styleguide/cppguide.html#Namespaces
  2: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using

- Changed "const std::string_view&" uses to "std::string_view" std::string_view is already a "view". It does not contain the data it points to. Thus, passing it by const reference is somewhat redundant and unecessary.

  This change vastly simplifies the opening logic in the Levenshtein function as the sting_views can simply be swapped if the second string is shorter, instead of creating new ones.

- Cast std::string_view::length() return value to int The default return type from std::string_view::length(). Signed <-> unsigned integer conversion should generally be avoided. Where it cannot easily be avoided (like here), conversion should be made explicit.

- Avoid raw for loops For initializing the first row, a raw for loop can be easily avoided in conjunction with C++20's for loop initializer syntax. Avoiding raw loops [1] can aid in avoiding off-by-one and other common errors.

  The first row initialization code has been converted to make use of the new ranged for loop syntax. The argument parsing and outer word loop has also been modified to use ranged for loops.

  For sake of leaving the core algorithm intact, the inner loops are still raw loops. Avoiding these would require converting to using iterators, which are somewhat less readable (IMHO).

  1: See Sean Parent, "C++ Seasoning" for details

- Removed "return 0" C++ main functions (unlike C) return 0 by default. The return statement can be ommited.

- No Abbreviations "curr" and "prev" were not universal [1] and in this case even caused shadowing of a function (std::prev). Further, "str1" and "str2" are indistinguishable from "s1" and "s2" in their meaning, abbreviate and introduce another problem, which is encoding their type name in the variable name - another "bad" practice.

  1: See Kate Gregory, "Naming is Hard: Let's do better"

Further possible work:
At the cost of performance (so not good for this "benchmark"), the code could be further simplified using std::ranges and std::views.
Use of std::cout could be converted to std::print[ln]. At this point in early 2025 compiler support for these is still evolving and may not be universally available.

I have:

* [ ] Read the project [README](../README.md), including the [benchmark descriptions](../README.md#available-benchmarks)
* [ ] Read the PR template instructions before I deleted them
* [ ] Understood that if I have changed something that could impact performance of one or more contributions, I should provide results benchmark runs, using the `run.sh` script, from before and after the change.


PEZ:
Hello,

It looks like valid updates to me. Maybe the of renaming the variables makes it a bit less easy to compare with the other implementations, but I agree there is also value in making well-visited code be representative for the coding standards in active use.

However, I think this really qualifies as:

> something that could impact performance of one or more contributions

So please provide benchmark runs showing that the performance is not impacted by the changes. You educated conjecture about it is not enough. We've seen very surprising changes in performance from tiny changes, like: https://x.com/ThePrimeagen/status/1878937668492648541

int2str:
I've benchmarked this before on my slow notebook and ensured no performance impact.
Running the benchmark on my fast machine however did result in a minor regression.

So I've adjusted this patch slightly by reverting that part of my change.

I've now also run the benchmark again with no statistically significant performance changes. See attached (hopefully).

Also note that C++ is significantly faster than C and others on all of my machines. So not sure what the last time was that you updated the visuals. Would be nice to it reflected that C++ is as fast, if not faster than C.

Also let me know if there's any interested in setting up GitHub actions so that performance tests run automatically so people have a "source of truth".

image

int2str:
Lol, nevermind the GitHub comment, just saw the automation exists :)

Issue #341, opened 2025-01-11
Start times: Make the programs do the benching too


PEZ:
As I mentioned in https://github.com/bddicken/languages/issues/336#issuecomment-2585330995 I think we can take inspiration from [Dave Plummers Eratosthenes drag-race benchmark](https://github.com/PlummersSoftwareLLC/Primes) in order to control when we want start-times to matter or not.

Let's use **levenshtein** as an example. What we are insterested in is how long takes for the program, given a bunch of strings, to figure out the minimum Levenshtein distance. So then we should only be measuring this. Not the start time of the executable. Not the time it takes to parse the input arguments and prepare for figuring out the minimum distance.

In Dave Plummer's drag-race each program is given the same amount of time, and they measure how many times the program can finish the task. This controls the length of the whole benchmark run much better and we can allow slow programs to participate without worrying that they will take forever. We can make it so that even R finishes one run of Levenshtein in, say, 15s . Then we give each program 30s to run, and allow them to use the same time to warm up their JIT engines. Even with 100 languages, slow or fast, it would take a maximum of 2 hours to run a benchmark. Say we have 5 benchmarks and 100 languages, it would take 10 hours.

I think we could have some easy-to-parse config file that specifies the benchmark. How long benchmark time that is allowed, the name of the input file, maybe something more. Then each program has some infrastructure to read and parse the config file, read and parse the input, run the measured function, measure how many times it runs, measure how long time it took, output the result of the function as well as the measurements.

It would be some considerable work to update the contributions, but since it's crowd sourced work, I think we can do it. I can do it for Java, C, and Clojure for reference and learning before we go ahead updating the rest. (My C is not up-to-date, nor up-to-speed, but I think I can do it, haha.)

igouy:
> So then we should only be measuring this.

Without the other stuff the programs won't figure out the minimum Levenshtein distance.

PEZ:
So the idea is some little “tooling” around the call to the thing being benchmarked. (As opposed to using tooling libraries of programs built for each platform.) It's going in-process to the extreme that hyperfine goes out-of-process.

For Clojure it may look like so:

```clojure
(defn benchmark [run-ms f]
  (let [t-start (System/currentTimeMillis)
        t-stop (+ t-start (* run-ms))
        rs (reduce (fn [results i]
                     (let [result (f) ; <- call the benchmarked function
                           t (System/currentTimeMillis)]
                       (if (< t t-stop)
                         (conj results [i t result])
[...]

Issue #336, opened 2025-01-10
Increase runtime of programs


bddicken:
Something that has been brought up repeatedly is the fact that this benchmark includes interpreter/jit/whatever startup times, which makes it "unfair" relative to purely compiled languages. While this is partly true, it is not unreasonable to have a benchmark that does include these startup times, as it is real cost in some scenarios.

One thing that could be done to both (A) help mitigate this problem and (B) uncover other performance differences is make the program run longer. This can be done by giving each larger input values or more data to process. 

At the time of writing this, there are four programs. Here's how we could change this for each:

- `loops`: update the implementations to take a variable amount of iterations. Then, pass in a value that causes longer runtime (perhaps 10 billion instead of 1, for example)
- `levenshtein` - Add more strings and longer strings to the passed-in input
- `fibonacci` - Call with 43 or 44 instead of 40
- `hello-world` nothing!

PEZ:
I don't think increasing runtime of programs will do enough to compensate for start times to carry the weight it adds. E.g. Julia has a start time of around 200ms on the M1 machine I run benchmarks on. It will continue to be heavily penalized even with 10X longer runtimes.

Some languages already take a lot of time running some of the tests. R takes 4 minutes on my M1 machine to do Levenshtein. And I run each program 15 times. So that's an hour just for R, just for Levenshtein. If it would take 10 hours I would run out of patience. 😄 

I am at where I think that the only way to control the start-times issue, without involving separate benchmark tooling per language/platform, is to require from the implementations that they do the measurements and report the timings as part of their output. This is how they do it in Dave Plummer's Eratosthenes drag-race. That way we can also avoid the problem with that parsing input gets to be part of the tests. (Separate parsing benchmarks can be designed to measure such things.)

igouy:
> as it is real cost in some scenarios

And there are many scenarios.

And what is done here will not apply to all scenarios.

And seemingly no statement of how applicable what has been done here, is intended to be to different scenarios.

Latency or Throughput?

igouy:
> without involving separate benchmark tooling per language/platform, is to require from the implementations that they do the measurements and report the timings

If "the implementations" must "do the measurements and report the timings" then would that not be "separate benchmark tooling per language/platform" ?

zierf:
> If "the implementations" must "do the measurements and report the timings" then would that not be "separate benchmark tooling per language/platform" ?

One simple thing would be taking the current time directly before and after invoking the algorithm in question.
[...]

PR #365, opened 2025-01-17
Programs do the benching

PEZ:
I have:

## Description of changes

This is the start of what I said in #341 that I should try to implement in-process benchmarking for Clojure, Java, and C. Starting with Clojure, to get it as right as possible before cloning the approach to any other languages.

I think the changes are best described in the **New runner** section in the updated [README.md](https://github.com/PEZ/languages/blob/programs-do-the-benching/README.md), plus:

* Loops [README](https://github.com/PEZ/languages/blob/programs-do-the-benching/README.md)
* Fibonacci [README](https://github.com/PEZ/languages/blob/programs-do-the-benching/fibonacci/README.md)
* Levenshtein [README](https://github.com/PEZ/languages/blob/programs-do-the-benching/levenshtein/README.md)

Addressing:

* #341
* #336


PEZ:
Now there are implementations for C, Clojure and Java, and Babashka (which can use the Clojure benchmarking utility). Here's what a levenshtein run looks like:

example-results-csv

int2str:
I'll have a crack at the C++ version of the benchmark.

PEZ:
Awesome, @int2str! I'm curious if you'll be able to use benchmark.c for it. Don't hesitate to make changes to it that would enable sharing.

int2str:
I've drafted a C++ version -> https://github.com/int2str/languages/commit/c5f3f027cfe79242923b79d5d387ec460e84162a

Haven't tried using the C version. I might try that as well.
Though it feels "icky" somehow :D lol

Also, if the return type for the benchmark ever is something more complex than an int, I'd prefer to have my template<> blanket to comfort me :D

PEZ:
> if the return type for the benchmark ever is something more complex than an int

It already is for Levenshtein where the return type is an array.

The reason for sharing is that it is less code to maintain. But if it makes things awkward for the CPP benchmarks that is a reason to implementing something more CPP friendly. Also, providing a separate CPP version could be good for reference, I think.

int2str:
Updated my draft to print dots as requested.

Btw. Maybe the warm-up runs shouldn't be the same amount as the final run? As in, why not do a 1 or 2 second warm-up and then the 10s run. Seems a bit much to do 10s for both, no?

PEZ:
I honestly don't know how much warmup is needed. But for JIT compilers it is different than for non-JIT. I can do a bunch of tests on Java and Clojure with different warmups and see if I can find a better guess than “same as benchmark”.

PEZ:
Can you file a PR on my fork with your changes, @int2str?

PEZ:
I'm closing this PR and will move the changes to a separate repository/project which will be focused on benchmarking with minimal impact from start/setup times: https://github.com/PEZ/languages-fun

Because the world simply don't have enough of microbenchmark projects! 😄 

@int2str PR with your C++ benchmark runner welcome there!

PEZ:
Alright @int2str. @bddicken and I decided to go for this runner in this repo, so I archived that other repo and now hope you'll PR your C++ implementation here. I'm sorry for the back-and-forth.

int2str:
Will do some time over the weekend.

No worries about the back-and-forth :)

PEZ:
Cool! I added a separate `warmup-ms` arg to the programs, instead of just using `run-ms` for it. The `run.ms` script defaults to `2000` for warm-up. We can tweak that later if we see that more is needed for some environment. I also changed the `benchmark.{clj,java,c}` utilities to not run at all if the `run-ms` is 0. And the `run.sh` script sends `1`, and `0`, for `run-ms` and `warmup-ns` respectively in the correctness test run. I kept it simple in `run.{clj,java,c}` and just don't look at the return value of the warmup run. E.g. Executing something like `./c/run 0 0 levenshtein-words.txt` will crash or not terminate or something. I think that's fine, running it with 0 `run-ms` doesn't make sense to me.

PR #367, opened 2025-01-18
CPP: Fibonnaci -> More idiomatic C++ code

int2str:
Changes from the C-based version:
- Added command line argument count check
- C++ does not require "return 0" in the main function
- Trailing return types
- Convert argv/argc to std::span to void pointer arithmetic and allow for simpler iteration (if needed)
- C++23 provides std::print() and std::println() to replace std::cout and friends
- C++23 ranges and algorithms to clarify intent and to avoid raw for loops. See Sean Parents "No raw for loops" and other sources.

Minor performance tweak:
- Replaced separate if() checks for n == 0 and n == 1 with single n < 2 check.

Note:
The performance tweak is also applicable to the C version.

Benchmark:
image

### DIFF
--- a/fibonacci/cpp/code.cpp
+++ b/fibonacci/cpp/code.cpp
@@ -1,24 +1,48 @@
-#include <iostream>
-using namespace std;
+//
+// "Modern", more idiomatic C++ version of the naive Fibonacci
+// sequence generator
+//
+// Changes from the C-based version:
+// - Added command line argument count check
+// - C++ does not require "return 0" in the main function
+// - Trailing return types
+// - Convert argv/argc to std::span to void pointer arithmetic
+//   and allow for simpler iteration (if needed)
+// - C++23 provides std::print() and std::println() to replace
+//   std::cout and friends
+// - C++23 ranges and algorithms to clarify intent and to avoid
+//   raw for loops.
+//   See Sean Parents "No raw for loops" and other sources.
+//
+// Minor performance tweak:
+// - Replaced separate if() checks for n == 0 and n == 1 with
+//   single n < 2 check.
+//
 
-int fibonacci(int n)
-{
-    if (n == 0)
-        return 0;
-    if (n == 1)
-        return 1;
-    return fibonacci(n - 1) + fibonacci(n - 2);
+#include <algorithm>
+#include <print>
+#include <ranges>
+#include <span>
+
+[[nodiscard]] auto fibonacci(int n) -> int {
+  if (n == 0)
+    return 0;
+  if (n == 1)
+    return 1;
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks!

> Replaced separate if() checks for n == 0 and n == 1 with single n < 2 check.

The C version is considered reference for this. So that means we rather want all contributions to do the two checks. At least until we decide to change the reference. This is not consistently applied so far, but anyway, please keep the two checks for now.

int2str:
Ok, I ran the tests again on my super slow notebook to take some measurement noise out of it. It's very close and certainly in line with the C implementation.

I'll leave it up to you guys whether a few ms regression is worth it to showcase the actual language differences vs. just using the same program with a different compiler. Not sure what's more important here TBH.

Happy to abandon, happy to have it merged now. Your call.

Thx either way.

image

fibonacci/cpp/code.cpp
```diff
@@ -1,24 +1,46 @@
...
-    if (n == 0)
-        return 0;
-    if (n == 1)
-        return 1;
+#include <algorithm>
+#include <print>
+#include <ranges>
+#include <span>
+
+[[nodiscard]] constexpr auto fibonacci(int n) -> int {
```
cyrusmsk, line 27: Why do you constexpr here?
What do you want to calculate at compile-time?
Do other langs use any compile-time mechanisms? Is it even allowed?

PEZ (changes requested):
I think we can't have `constexpr` in there.

PEZ:
> Happy to abandon, happy to have it merged now. Your call.

Except for `constexprr` I'm good with merging all right. You're the CPP expert of us two, and have a bigger say, I would say. 😄

int2str:
[[constexpr]] is a suggestion to the compiler, not a rule. It means this function *can* be invoked in a constexpr context, not that it *will*.

In this example, since the arguments to the function are a command line argument, it clearly cannot be evaluated at compile time. Further more, the before and after benchmark results show that this has no effect.

Again, just showcasing language features here.
But also happy to remove....

Just let me know.

int2str:
Removed constexpr.

fibonacci/cpp/code.cpp
int2str: Explanation below, but removed it just to remove any doubt.

int2str:
New comparative C/CPP benchmark after removing 'constexpr'

image

PEZ:
Thanks for improving this project, @int2str! 🙏 And thanks for removing `constexpr`, even though it didn't break any rules after all. It's cleaner this way and we won't have to have _that_ discussion at least. There is no lack of issues to discuss, after all.

PR #368, opened 2025-01-18
CPP: Hello, World! C++23 style

int2str:
- C++ does not require "return 0" in main()

- std::endl should almost never be used as it flushes the output in addition to adding a newline. 
This is a performance issue in real world programs and should not be "trained" by default. If iostream is still used, simply print newlines like this: std::cout << "Hello, World!\n";

- Move to C++23 std::print / std::println

- Trailing return types

### DIFF
--- a/hello-world/cpp/code.cpp
+++ b/hello-world/cpp/code.cpp
@@ -1,6 +1,5 @@
-#include <iostream>
+#include <print>
 
-int main() {
-    std::cout << "Hello, World!" << std::endl;
-    return 0;
+auto main() -> int {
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks!

Please also provide benchmark runs so that we can see that we're not degrading anything.

int2str:
For Hello, World, this will be well into the measurement noise, but sure...

image

PEZ:
Thanks. 🙏  Very much into the measurement noise territory! 😄 It's more so that I don't have to pull the changes and run this myself to see that things still compile and run.

PR #369, opened 2025-01-23
add vectorized python

notlober:
```
Benchmarking Python Vectorized
Benchmark 1: python ./py-vec/code.py 40
  Time (mean ± σ):      18.2 ms ±   0.1 ms    [User: 13.1 ms, System: 4.1 ms]
  Range (min … max):    18.2 ms …  18.3 ms    3 runs

```
```
Benchmarking CPP
Benchmark 1: ./cpp/code 40
  Time (mean ± σ):     637.3 ms ±   0.6 ms    [User: 632.5 ms, System: 3.0 ms]
  Range (min … max):   636.9 ms … 637.9 ms    3 runs
```

I fully vectorized the operation using math and implemented it in python, currently it gets around 18 ms in my machine compared to c++'s 640 ms.

enjoy 30x speed boost over previous best.

proof of equivalence:
```
import sys
import random


def sum_j_mod_u(N, u):
    # N mod u:
    L = N % u
    # Floor(N / u):
    m = N // u
    # sum_{k=0 to u-1} k = u*(u-1)//2
    # sum_{k=0 to L-1} k = L*(L-1)//2
    return m * (u*(u-1)//2) + (L*(L-1)//2)

def main():
    u = int(sys.argv[1])
    r = random.randint(0, 10000)
    total = sum_j_mod_u(100000, u) + r
    print(total)  # since a[r] == total
    for i in range(10000):  # 10k outer loop iterations
        for j in range(100000):  # 100k inner loop iterations, per outer loop iteration
            a[i] += j % u  # Simple sum
        a[i] += r  # Add a random value to each element in array
    print(a[r])  # Print out a single element from the array


main()
```


PEZ:
No hand-vectorized solutions accepted here.

PR #370, opened 2025-01-25
In-process benchmarking

PEZ:
Couldn't reopen #365, so a new PR. 😄 

I have:

## Description of changes

This is the start of what I said in #341 that I should try to implement in-process benchmarking for Clojure, Java, and C. Starting with Clojure, to get it as right as possible before cloning the approach to any other languages.

I think the changes are best described in the **New runner** section in the updated [README.md](https://github.com/PEZ/languages/blob/programs-do-the-benching/README.md), plus:

* Loops [README](https://github.com/PEZ/languages/blob/in-process-bench/README.md)
* Fibonacci [README](https://github.com/PEZ/languages/blob/in-process-bench/fibonacci/README.md)
* Levenshtein [README](https://github.com/PEZ/languages/blob/in-process-bench/levenshtein/README.md)

Addressing:

* #341
* #336


README.md
```diff
@@ -1,76 +1,207 @@
...
+Because of the above, each language will have to have some minimal utility/tooling for running the function-under-benchmark as many times as a timeout allows, plus reporting the measurements and the result. Here are three implementations, that we can regard as being reference:
+
+* [benchmark.clj](lib/clojure/src/languages/benchmark.clj)
+* [benchmark.java](lib/java/languages/Benchmark.java)
+* [benchmark.c](lib/c/benchmark.c) (This one may need some scrutiny from C experts before we fully label it as *reference*.)
+
+You'll see that the `benchmark/run` function takes two arguments:
+
+1. `f`: A function (a thunk)
+1. `run-ms`: A total time in milliseconds within which the function should be run as many times as possible
```
benknoble, line 53: Please make sure you're using a monotonic clock for these measurements (wall clock time can jump around).

I don't know how this actually works in other languages, but in Racket you can provide a time-limit for a body of execution (see [`call-with-time-limit`](https://docs.racket-lang.org/reference/Sandboxed_Evaluation.html#%28def._%28%28lib._racket%2Fsandbox..rkt%29._call-with-deep-time-limit%29%29)).

PEZ: Good point. I made sure it was a monotonic clock for Java and Clojure, but I shall now check the other language implementations. It's probably fine, but I didn't pay attention to that so could have slipped in some wall clocking.

PR #393, opened 2025-02-01
Port Crystal to in-process benchmark

### LINKED ISSUE #387: Crystal In-process runner
- [ ] Benchmark utility (Maybe it can share with C#? #381)
- [ ] loops
- [ ] fibonacci
- [ ] levenshtein
- [ ] hello-world

Tamnac:
Closes #387

I have:

## Description of changes

Uses a benchmark function in lib/crystal which takes in a time and block to bench

I also added hello world, cause crystal didn't have it


lib/crystal/benchmark.cr
```diff
@@ -0,0 +1,25 @@
...
+	while(times.sum < run_ms && !(times.sum == 0 && times.size > 0))
+		a = Time.measure {result = yield}
+		times << a.milliseconds
+	end
+
+	# mean_ms,std-dev-ms,min_ms,max_ms,times,result
+	unless times.empty? || warmup?
+		puts "#{times.sum / times.size},#{times.std_dev},#{times.min},#{times.max},#{times.size},#{result}"
+	end
+end
```
PEZ, line 25: There probably should be two functions in the interface here: `bench`/`run`, and `format_results`. Where the run function returns its results (the `times` array in this case) and `format_results,` well, formats a given result.

This way a run module like that for levenshtein can sum up the distances returned in the result and get help to format the output.

levenshtein/crystal/leven.cr
PEZ, line 1: We want to be able to find all Levenshtein implementation by globbing in the shell, like `levenshtein/*/levenshtein.*`, so we're naming them all `levenshtein.`, in this case it should be `levenshtein.cr` then.

fibonacci/crystal/fib.cr
PEZ, line 1: Same as below with naming the file. Please use `fibonacci.cr`.

levenshtein/crystal/leven.cr
```diff
@@ -0,0 +1,50 @@
+def levenshtein(strings : Array(String))
+	# stolen from zig - reuse bufs
+	max_len = strings.map(&.size).max
+	buf1 = Array(Int32).new(max_len+1, 0)
+	buf2 = Array(Int32).new(max_len+1, 0)
+
+	strings.each_combination(2).sum do |pair|
+		levenshtein_distance(pair[0], pair[1], buf1, buf2)
+	end
+end
```
PEZ, line 10: This function should return an array with the distances. And the run module should sum them up. I bet that it doesn't matter a ton, but the sum is only for correctness check, and we're trying to get all languages to do it in the same way.

lib/crystal/benchmark.cr
PEZ: There should be some status reporting here. We do it super simple with printing a dot every second. Se other benchmark modules for reference. (Clojure is most similar to Crystal of the current implementations, so may be the easiest one to port.)

PEZ (changes requested):
Thanks for contributing! 🙏 Please also provide output from running the benchmark with Crystal.

levenshtein/crystal/leven.cr
zierf: I just noticed that buffers are created here before comparing all the strings. This means that they don't have to be recreated within each individual comparison, but only once per benchmark. That's roughly how I would like to change it for Rust and Zig (see [this discussion](https://github.com/bddicken/languages/pull/386#discussion_r1938322989)).

I think adding it up in the benchmark itself makes a difference. You could just add up a small integer after each run, which is a very simple and fast operation. Having to output a complete list of results, on the other hand, requires reserving a list of unknown size and can't be done at compile time, which may force the use of the heap (at least without VLAs) for some languages.

PEZ: For the languages I have ported to the new runner it didn't make a measurable effect. Which may speak for that it doesn't matter, but I think there is value in that all languages are required to do it the same way. And to be allowed to do it as efficiently as they can. It isn't strictly part of what we want to benchmark, but rather a requirement for the sake of being able to do a correctness check.

It isn't a ton of words, so the array should be small. Will it risk forcing the use of the heap anyway, you think, @zierf? We could define a max for the number of words, and then the implementations can allocate it at compile time?

PEZ (approved):

PEZ:
Thanks, @Tamnac ! 🙏 ❤️

levenshtein/crystal/leven.cr
zierf: I would rather not specify an exact or maximum number of words, that seems to make more sense in a real-world scenario.

The length and number of words in the input file should be able to change at any time without making the programs invalid.

If you wanted to be able to specify everything at compile time, you could just write the desired words into a fixed array. Then you could use `comptime` in Zig to have the compiler preprocess everything in advance and at the end the program would just `return 554324;`. Perhaps a little exaggerated, but then the comparison between languages ​​no longer makes sense.

Maybe a bit exaggerated, but the benchmark should measure somewhat useful. And if you read something from a file, you should also make sure that you always provide enough buffer and the number of words or their lengths can change anytime.

Of course, we don't need to overdo it, so that niche languages ​​are not pushed to absurd limits.

PR #401, opened 2025-02-05
Fortran modularize w retained performance

PEZ:
I have:

## Description of changes

The modularization I did on the Fortran fibonacci program made the compiler drop the optimization that made the program totally kill all other implementations. With [the help of o3-mini-high](https://chatgpt.com/share/67a368c3-bea0-8003-b8dd-16e1757eb083) I tracked it down to that I had made the function “public” which prevented some optimization. So now I have made the function private inside the module and expose a public function only for indirection. This reclaimed the performance.

**Before, ~25ms:**

```
❯ ../run.sh -u PEZ -l Fortran
Results will be written to: /tmp/languages-benchmark/2025-02-05T13-01-48Z_PEZ_10000_42fc727_Fortran.csv

Running fibonacci benchmark...

Checking fibonacci Fortran
./fortran/run 1 0 37
Check passed
Benchmarking fibonacci Fortran
./fortran/run 10000 2000 37
..
..........
fibonacci,2025-02-05T13:01:48Z,42fc727,true,PEZ,Apple M4 Max,64GB,darwin24,arm64,gcc (Homebrew GCC 14.2.0_1) 14.2.0,g++ (Homebrew GCC 14.2.0_1) 14.2.0,19.1.6,Homebrew clang version 19.1.6,Fortran,10000,24.738094,.440168,23.694000,26.291000,405

Done running fibonacci benchmark
Results were written to: /tmp/languages-benchmark/2025-02-05T13-01-48Z_PEZ_10000_42fc727_Fortran.csv

```

**After, ~0.1ms:**

```
~/Projects/languages/fibonacci  main[!✔] 
❯ ../run.sh -u PEZ -l Fortran
Results will be written to: /tmp/languages-benchmark/2025-02-05T13-02-27Z_PEZ_10000_42fc727_Fortran.csv

Running fibonacci benchmark...

Checking fibonacci Fortran
./fortran/run 1 0 37
Check passed
Benchmarking fibonacci Fortran
./fortran/run 10000 2000 37
..
..........
fibonacci,2025-02-05T13:02:27Z,42fc727,true,PEZ,Apple M4 Max,64GB,darwin24,arm64,gcc (Homebrew GCC 14.2.0_1) 14.2.0,g++ (Homebrew GCC 14.2.0_1) 14.2.0,19.1.6,Homebrew clang version 19.1.6,Fortran,10000,.127350,.026440,.103000,2.110000,78524

Done running fibonacci benchmark
Results were written to: /tmp/languages-benchmark/2025-02-05T13-02-27Z_PEZ_10000_42fc727_Fortran.csv
```


zierf:
I'm confused, how can Fortran complete the benchmark so absurdly quickly compared to all other languages? It almost looks as if it already determines the desired result at compile time. Which it shouldn't be able to do, since the desired number is only given as a parameter at runtime.

For me it manages maybe just over 217-230 runs in C/Rust/Zig/Cpp, which is still a bit far from your initial 405 runs on the M4 Max, but 78,524 times seems a bit high to me, especially for so many recursive calls.

I've seen some [integer(8)](https://github.com/bddicken/languages/blob/main/fibonacci/fortran/fibonacci.f90#L10) in the code. Unfortunately, I'm not really familiar with Fortran, but if that is the intended integer size (signed?) for the number with the result, then the expected number of 24,157,817 as a Fibonacci number of 37 can never fit in there (nor can a `u16`). Shouldn't actually pass the check then either. 🤔

PEZ:
gfortran seems to figure out that the recursion will run the function for low `n` very often, and replaces those calls with precomputed values: https://godbolt.org/z/oqorn5dvj

But only if the fibonacci function is called from the same module, without that: https://godbolt.org/z/xoWh95nGj

zierf:
It looked a lot like pre-calculation. I even loaded it into Godbolt briefly, but I'm not too familiar with it.

Apparently _Nim_ also carries out [a similar optimization](https://forum.nim-lang.org/t/4253#26501) in some cases and with a little extra help they managed to trigger the optimization for _gcc/g++_ too.

_Clang_ should also be able to [optimise the fibonacci test](https://github.com/JuliaLang/Microbenchmarks/issues/29). Unfortunately, I couldn't get it to work with either compiler, _Clang_ just runs much slower than _gcc/g++_.

A bit surprising, since Fortran is compiled here with _gfortran_ and, like the C(pp) compilers, this is a GNU program. Strange that _gfortran_ leaves its siblings so far behind here.

PEZ:
Yes, I wonder a lot how it manages to do that. And if there are conditions when gcc would do it with C code.

PR #409, opened 2025-02-07
Racket in process

PEZ:
I have:

## Description of changes

Ported the Racket contributions to the in-process runner.

According to ChatGPT there is no built-in nanoseconds monotonic clock, so we use a millisecond one. Works for the current benchmarks, at least.

@benknoble, I'll merge this PR as soon as I have benchmark results. But I don't know Racket so may be doing some stupid things. If you see something, please consider filing a PR fixing. 🙏 

A benchmark run: https://gist.github.com/PEZ/d78723c993744027a2fe3e39d4283db2

### DIFF
--- a/fibonacci/racket/fibonacci.rkt
+++ b/fibonacci/racket/fibonacci.rkt
@@ -1,0 +1,18 @@
+#lang racket/base
+
+(require racket/require
+         (for-syntax racket/base) ;; filtered-in fails without this
+         (filtered-in
+          (λ (name)
+            (and (regexp-match #rx"^unsafe-fx" name)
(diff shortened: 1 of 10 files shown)


benknoble:
I think there are some things worth fixing, but I'll point out [`current-inexact-monotonoic-nanoseconds`](https://docs.racket-lang.org/reference/time.html#%28def._%28%28quote._~23~25kernel%29._current-inexact-monotonic-milliseconds%29%29). You can also get a [`nanosecond` version on supported platforms](https://docs.racket-lang.org/monotonic/index.html#%28def._%28%28lib._monotonic%2Fmain..rkt%29._current-monotonic-nanoseconds%29%29), though that introduces a dependency.

PR #417, opened 2025-02-09
Crystal: don't preallocate bufs to comply with new requirements

Tamnac:
I have:

## Description of changes

We discussed this in the zig pr. Runs a little slower, as expected

before
image

after
image


PEZ:
🙏

PR #426, opened 2025-04-10
Update code.php

oceanwap:
I have:

## Description of changes

PHP JIT doesn't work so well for global code, wrapping it in a function makes JIT work better and code faster.
Also in a real word project a code would be wrapped in a class or function, so it's more realistic scenario.

### DIFF
--- a/loops/php/code.php
+++ b/loops/php/code.php
@@ -1,12 +1,18 @@
 <?php
 
 $u = (int) $argv[1];                                # Get an input number from the command line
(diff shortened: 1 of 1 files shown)


oceanwap:
@bddicken Please review the PR

jszczypk:
I have got almost 40% shorter time with JIT by wrapping the code in function.

oceanwap:
@bddicken 
I am not sure who will merge it, can someone mere it?

